PR #388, opened 2026-02-11
Phase 4: Instance-side class var access (BT-413)

jamesc:
## Summary

Implements instance-side class variable access via `self class` message chain (ADR 0013 Phase 4).

**Linear issue:** https://linear.app/beamtalk/issue/BT-413

### Codegen fix
- Fixed keyword colon handling in class method self-sends for built-in exports (`new:` → `new`, matching module function exports)
- This was a pre-existing bug that prevented value type class methods from calling `self new: args`

### Test fixtures
- **`InstanceAccessCounter`** (Actor): class variable `instanceCount`, class methods `create`/`instanceCount`, instance method `totalCounters` using `self class instanceCount`
- **`ClassVarPoint`** (Object subclass): same pattern for value types, proving both Actor and value type instances work

### Stdlib tests (16 new assertions)
- `instance_class_var_access.bt`: Actor instance reads class variable via `self class instanceCount` chain
- `instance_class_var_access_value_type.bt`: Value type instance reads class variable via same chain

## How it works

The feature is compositional — it falls out of Phase 3 (BT-412):
1. `self class` → `class_of_object/1` → returns `#beamtalk_object{}` class object
2. `(self class) instanceCount` → `beamtalk_message_dispatch:send` → detects class object → `class_send/3` → invokes class method
3. Class method reads class variable from gen_server state


jamesc:
Addressed Copilot review: added explicit `spawnWith:` → `spawn` mapping for class method self-sends (d863310). CI passes.

PR #465, opened 2026-02-13
Add ETS class hierarchy table for reliable method reflection (BT-510)

jamesc:
## Summary

Fixes [BT-510](https://linear.app/beamtalk/issue/BT-510/fix-dynamic-class-methods-reflection-returning-empty-list): `Counter methods` was returning empty/incomplete lists for user-defined classes because `build_flattened_methods/4` silently fell back to local-only methods when `whereis_class(SuperclassName)` returned `undefined` during `init/1`.

## Approach: ETS Class Hierarchy Table

Added a `beamtalk_class_hierarchy` ETS table storing `{ClassName, SuperclassName | none}` pairs. This eliminates timing dependencies in hierarchy queries.

### Key changes

- **`ensure_hierarchy_table/0`** — Creates ETS table at app startup (owned by app master, survives class process crashes). TOCTOU-safe with try/catch.
- **`init/1`** — Inserts `{ClassName, Superclass}` into ETS on class registration (immediately visible to all processes).
- **`inherits_from/2`** — Rewritten from gen_server call chain to pure ETS reads. O(1) per hierarchy level, no process messaging.
- **`build_flattened_methods/4`** — Added debug log on fallback path (superclass unavailable during init).
- **`beamtalk_runtime_app.erl`** — Calls `ensure_hierarchy_table()` at startup so table is owned by application master.

### Tests

- 5 new unit test suites: `hierarchy_ets_populated`, `hierarchy_inherits_from`, `hierarchy_methods_inherited`, `hierarchy_methods_normal_order`, `hierarchy_orphan_registration`
- New `tests/stdlib/class_reflection.bt` — verifies Counter inherits Object methods
- Enhanced `tests/e2e/cases/class_reflection.bt` — 20 new inherited method assertions

### Note

The existing `invalidate_subclass_flattened_tables` broadcast mechanism already handled out-of-order registration. The ETS table is defense-in-depth for `inherits_from` and enables future targeted invalidation.

## Summary by CodeRabbit

* **New Features**
  * Improved class hierarchy lookup performance with optimized registry system.
  * Enhanced class reflection to accurately display both local and inherited methods.

* **Tests**
  * Added comprehensive tests for class hierarchy registration and method inheritance.
  * Extended class reflection tests to verify method discovery across inheritance chains.

### DIFF
--- a/runtime/apps/beamtalk_runtime/test/beamtalk_object_class_tests.erl
+++ b/runtime/apps/beamtalk_runtime/test/beamtalk_object_class_tests.erl
@@ -939,11 +939,17 @@ hierarchy_methods_inherited_test_() ->
                          instance_methods => #{parentMethod => #{arity => 0}}
                      }),
 
-                     %% Synchronous barrier to ensure child processed rebuild
-                     _ = beamtalk_object_class:superclass(ChildPid),
-
-                     %% After parent registered — inherited methods included
-                     MethodsAfter = beamtalk_object_class:methods(ChildPid),
+                     %% Wait until rebuild is processed (bounded retry)
+                     WaitForInherited = fun Wait(0) ->
+                                                ?assert(false);
+                                            Wait(N) ->
+                                                M = beamtalk_object_class:methods(ChildPid),
+                                                case lists:member(parentMethod, M) of
+                                                    true -> M;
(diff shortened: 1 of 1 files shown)


jamesc:
@coderabbitai review

PR #488, opened 2026-02-14
Fix fuzz CI OOM/timeout detection and parser infinite loop on unclosed maps

jamesc:
## Summary

The nightly fuzzing workflow always reported success despite every run hitting OOM. This PR fixes both the CI detection and the underlying parser bug.

### Changes

**Workflow (`.github/workflows/fuzz.yml`):**
- "Check for crashes" now fails on `oom-*` and `timeout-*` artifacts, not just `crash-*`
- "Create issue on failure" creates GitHub issues for all failure types with specific details

**Parser (`expressions.rs`):**
- Added EOF guard in `parse_map_literal` loop to prevent infinite loop when error recovery reaches end of input on unclosed map literals like `#{key value`

**Tests (`mod.rs`):**
- `unclosed_map_literal_at_eof_terminates` — minimal repro of the fuzz artifact
- `unclosed_map_literal_bare_eof_terminates` — variant without surrounding parens

### Root Cause

The fuzzer produced input containing `#{key ...` (no `=>`, no `}`) at EOF. The `parse_map_literal` loop's error recovery consumed tokens to EOF, then `continue`d back to the top of the loop. At EOF, `advance()` is a no-op, creating an infinite loop that consumed memory until OOM at 4GB.

All other parser loops already had EOF guards (`parse_block`, `parse_match_expression`) or break on the error path (`parse_list_literal`).

### Verification

- All CI checks pass (711 core tests, 291 compiler tests, 1220 runtime tests, 1305 stdlib tests, E2E)
- Clippy clean, fmt clean, dialyzer clean

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed parser hang when encountering unclosed map literals at end of file.

* **Tests**
  * Added regression tests to ensure unclosed map literal parsing terminates with an error.

* **Chores**
  * Improved fuzz workflow failure handling and reporting: aggregated failure detection (crash, OOM, timeout), consolidated failure status, and enhanced issue creation with multi-type titles and detailed run summaries.

### DIFF
--- a/.github/workflows/fuzz.yml
+++ b/.github/workflows/fuzz.yml
@@ -88,6 +88,12 @@ jobs:
             FAILED=1
           fi
 
+          # Fail if fuzz step failed but produced no artifacts (infra error, cargo-fuzz crash, etc.)
+          if [ $FAILED -eq 0 ]; then
+            echo "❌ Fuzz step failed but no artifacts were produced; failing to avoid false green."
+            FAILED=1
+          fi
+
           exit $FAILED
 
       - name: Create issue on failure


jamesc:
Addressed CodeRabbit review comment in ca6797d:

3. **False green on no artifacts**: Added guard that fails when fuzz step failed but produced no artifacts (infra error, cargo-fuzz crash, etc.) ✅

All 3 review threads now resolved.

PR #516, opened 2026-02-14
Parser: support comma (,) as string concatenation binary operator (BT-415)

jamesc:
## Summary

Adds support for `,` as a binary operator for string concatenation, aliasing `++`. This matches the Smalltalk convention where `,` is a standard binary message on `Collection` (and `String`) for concatenation.

**Linear issue:** https://linear.app/beamtalk/issue/BT-415

### Parser
- Added "," to `binary_binding_power()` at precedence level 30 (same as `++`, `+`, `-`)
- Added `in_collection_literal` flag to `Parser` struct to distinguish comma-as-separator (inside `#()` and `#{}`) from comma-as-operator
- Blocks and parenthesized expressions reset the flag so comma works as an operator inside nested contexts

### Codegen
- Route comma operator to `generate_concat_op()` (same code path as `++`)
- Added defensive arity check for comma binary selector

### Tests
- Parser unit tests: comma as binary operator, comma as list separator, comma in block inside list
- Stdlib test: `'hello' , ' world' // => hello world`

### Documentation
- Updated `beamtalk-language-features.md` operator precedence table
- Updated `beamtalk-syntax-rationale.md` precedence list

### Snapshots
- Updated 8 compiler snapshots affected by improved comma handling in error recovery

## Release Notes

* **New Features**
  * Comma operator (`,`) now supported as a string and list concatenation operator, serving as an alias for the `++` operator.

* **Documentation**
  * Updated language documentation and operator precedence guide to reflect comma operator support for concatenation operations.

### DIFF
--- a/crates/beamtalk-core/src/source_analysis/parser/expressions.rs
+++ b/crates/beamtalk-core/src/source_analysis/parser/expressions.rs
@@ -547,6 +547,9 @@ impl Parser {
         }
 
         // Parse block body — statements separated by periods or newlines (BT-360)
+        // BT-415: Reset in_collection_literal so comma works as binary operator inside blocks
+        let prev_in_collection = self.in_collection_literal;
+        self.in_collection_literal = false;
         let mut body = Vec::new();
         while !self.check(&TokenKind::RightBracket) && !self.is_at_end() {
             let expr = self.parse_expression();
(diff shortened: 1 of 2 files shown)


jamesc:
Closing: after discussion, decided not to ship comma as a binary operator.

**Rationale:**
1. **Parser complexity** — requires `in_collection_literal` flag threading through 5 parser functions, with fragile save/restore on every exit path
2. **Future type checking** — comma would need special handling in the type checker too, adding complexity for a synonym of `++`
3. **Codegen coupling** — `,` is hardcoded to `generate_concat_op` rather than going through normal method dispatch, which fights the eventual type-checked architecture
4. **Marginal value** — saves 1 character vs `++`, which is already idiomatic and universally used

The `, other => self ++ other` method definition can remain in `String.bt` as a runtime-dispatched method if needed, without parser-level operator support.

PR #762, opened 2026-02-21
fix: non-local return (^) inside blocks in Actor methods (BT-761)

jamesc:
## Summary

Fixes BT-761: Non-local return (`^`) inside block arguments was broken in Actor subclass methods. A `^` inside a block (e.g. `ifTrue: [^"early"]`) now correctly causes an early return from the enclosing actor method, producing the proper `{reply, Value, State}` gen_server tuple.

## Approach

Applied the same try/catch NLR pattern from BT-754 (value type methods) to actor method codegen:

1. **NLR detection** — `expr_has_block_nlr()` scans the method body for `^` inside block closures
2. **Token setup** — A unique `make_ref()` token scopes the NLR to the current method
3. **try/catch wrapping** — The method body is wrapped to catch `{'$bt_nlr', Token, Value}` throws and produce `{reply, Value, State}`
4. **Re-raise non-NLR** — Unrelated exceptions pass through via `raw_raise`

Both actor dispatch paths are covered:
- **Class-based methods** (`methods.rs` — `generate_method_dispatch()`)
- **Legacy block-based methods** (`dispatch.rs`)

A shared `wrap_actor_nlr()` helper eliminates duplication between the two paths.

## Test Coverage

5 fixture files covering:
- `ifTrue:` with `^` (basic case)
- `ifTrue:ifFalse:` with `^` in both branches
- Local variable + conditional `^`
- `^` with computed value (`n * 2`)
- `ifFalse:` with `^`

18 test assertions in `tests/stdlib/actor_non_local_return.bt`, all passing.

## Design Notes

- **State on NLR**: The catch clause uses the *initial* `State` (entry state), so field mutations before `^` are not preserved. This matches Smalltalk semantics — `^` is designed for guard-clause early exits before state mutation.
- **Recursive actor methods**: Not supported with NLR (gen_server self-call would deadlock). This is a fundamental actor model constraint, not a bug.
- **`on:do:` compatibility**: The exception handler in `exception_handling.rs` explicitly re-raises `$bt_nlr` throws, so NLR works correctly through try/catch blocks in user code.

## CI

All checks pass: build, clippy, fmt, dialyzer, test, test-stdlib (1899 tests), test-e2e, runtime tests, integration tests, MCP tests.

## Summary by CodeRabbit

* **New Features**
  * Non-local returns within block closures in actor methods now properly propagate to the enclosing method, enabling safe early exits from blocks.

* **Tests**
  * Added comprehensive end-to-end test coverage for non-local return behavior in actor methods, including basic conditionals, branches, computed values, and local variable scenarios.

### DIFF
--- a/crates/beamtalk-core/src/codegen/core_erlang/gen_server/dispatch.rs
+++ b/crates/beamtalk-core/src/codegen/core_erlang/gen_server/dispatch.rs
@@ -267,26 +267,7 @@ impl CoreErlangGenerator {
 
                     // BT-761: Wrap in try/catch for non-local returns from block closures
                     let body_str = if let Some(token_var) = self.current_nlr_token.take() {
-                        let result_var = self.fresh_temp_var("NlrResult");
-                        let cls_var = self.fresh_temp_var("NlrCls");
-                        let err_var = self.fresh_temp_var("NlrErr");
-                        let stk_var = self.fresh_temp_var("NlrStk");
-                        let ctk_var = self.fresh_temp_var("CatchTok");
-                        let val_var = self.fresh_temp_var("NlrVal");
-                        let ot_pair_var = self.fresh_temp_var("OtherPair");
-                        format!(
-                            "let {token_var} = call 'erlang':'make_ref'() in\n\
-                             try\n  {body_str}\n\
-                             of {result_var} -> {result_var}\n\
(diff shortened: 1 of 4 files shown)


jamesc:
Duplicate of #760 — closing.

PR #882, opened 2026-02-24
Dynamic class creation: closure-based dispatch (Path 2) + BUnit/E2E tests (BT-838)

jamesc:
## Summary

Implements ADR 0038 Phase 4: closure-based dynamic class creation via ClassBuilder without a compiler round-trip. This is the capability that unblocks REPL self-hosting.

**Linear issue:** https://linear.app/beamtalk/issue/BT-838

### Key Changes

- **Path 2 detection** in `beamtalk_class_builder:register/1`: when `moduleName` is absent, routes to dynamic path with closure wrapping and field accessor generation
- **Closure wrapping** (`wrap_method_for_dispatch/1`): user closures (arity 0/1/2) wrapped to arity-3 dispatch protocol `fun(Self, Args, State) -> {reply, Result, State}`
- **Field accessors**: auto-generated getters for each field in dynamic classes
- **Hierarchy fallback** in `beamtalk_dynamic_object:dispatch/4`: falls back to `beamtalk_dispatch:super/5` for inherited methods (class, respondsTo:, printString, etc.)
- **Field defaults**: `addField:default:` stored in class state, applied at instantiation (including partial init via `new:`)
- **ClassBuilder.bt fix**: all 7 state mutations corrected to use `self.` prefix for proper codegen
- **Error handling**: `extract_error_kind` handles `{future_rejected, Reason}` for BUnit `should:raise:` assertions

### Tests

- **BUnit** (`stdlib/test/class_builder_test.bt`): 7 tests — create+send, field defaults, self-access, superclass, canUnderstand, sealed rejection, missing parameter
- **E2E** (`tests/e2e/cases/class_builder_dynamic.bt`): dynamic class lifecycle in REPL — create, dispatch, field access, reflection, error handling, cleanup

## Test plan

- [x] `just ci` passes (all checks, all tests)
- [x] BUnit: 7/7 pass
- [x] E2E: all pass
- [x] No regressions in existing tests

## Summary by CodeRabbit

* **New Features**
  * Two-path class model: compiled and dynamic classes coexist; dynamic classes support runtime closure-based methods, arity-3 dispatch, generated field accessors, and per-class default field values

* **Bug Fixes**
  * Method lookup now falls back to class hierarchy before doesNotUnderstand handling
  * Improved builder/instantiation error handling and stricter actor lifecycle error reporting

* **Tests**
  * Added unit and end-to-end tests for dynamic class creation, defaults, inheritance, dispatch, and lifecycle behaviors

### DIFF
--- a/stdlib/test/class_builder_test.bt
+++ b/stdlib/test/class_builder_test.bt
@@ -5,37 +5,39 @@
 //
 // Tests closure-based dynamic class creation through the ClassBuilder protocol.
 // Note: ClassBuilder and dynamic objects are Actors, so results must be awaited.
+//
+// tearDown removes any dynamically registered class, ensuring cleanup runs even
+// when assertions fail mid-test.
 
 TestCase subclass: ClassBuilderTest
+  state: cls = nil
+
+  tearDown =>
+    self.cls isNil ifFalse: [self.cls removeFromSystem]
 
   testDynamicClassCreateAndSend =>
-    _cls := (Object classBuilder name: #CBTestGreeter; addMethod: #greet body: ["Hello!"]; register) await.
+    self.cls := (Object classBuilder name: #CBTestGreeter; addMethod: #greet body: ["Hello!"]; register) await.
     obj := CBTestGreeter new.
-    self assert: (obj greet await) equals: "Hello!".
-    CBTestGreeter removeFromSystem
+    self assert: (obj greet await) equals: "Hello!"
 
   testDynamicFieldDefault =>
-    _cls := (Object classBuilder name: #CBTestFieldDog; addField: #name default: "Rex"; register) await.
+    self.cls := (Object classBuilder name: #CBTestFieldDog; addField: #name default: "Rex"; register) await.
     obj := CBTestFieldDog new.
-    self assert: (obj name await) equals: "Rex".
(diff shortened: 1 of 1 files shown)


jamesc (commented):
## CodeRabbit review response

### ✅ Fixed: `update_class` not refreshing `field_defaults` (commit dbba5c0b)

Added `field_defaults` to the `update_class` state update so hot-reloaded dynamic classes pick up new defaults immediately.

### 🚫 Dismissed: Arity-3 wrapping in `wrap_method_for_dispatch`

The arity-3 shortcut (`Fun` passed through unchanged) is intentional. Arity-3 IS the dispatch protocol: `fun(Self, Args, State) -> {reply, Result, State}`. This is used by `create_subclass` which generates closures already in the protocol format. Wrapping a protocol-conformant closure would double-wrap it and break the return tuple semantics.

### ⚠️ Noted: Test cleanup not guarded against assertion failures

The concern is valid. However, Beamtalk's BUnit TestCase currently has no `tearDown` mechanism that executes after assertion failures. Adding tearDown support would be a separate feature (tracked in a follow-up issue if needed). The test names use unique class name prefixes (`CBTest*`) to avoid collisions in case cleanup is skipped.

PR #997, opened 2026-02-27
Stdlib migration: reclassify existing classes to Value/Actor model (BT-925)

jamesc:
## Summary

Reclassifies 11 stdlib classes from `Object subclass:` to `Value subclass:` as part of ADR 0042 (Immutable Value Objects), Stage 3.

**Linear issue:** https://linear.app/beamtalk/issue/BT-925/stdlib-migration-reclassify-existing-classes-to-valueactor-model

## Key Changes

- **Abstract hierarchy roots** changed to `Value subclass:`: Collection, Number, Boolean, Exception
- **Concrete leaf classes** changed to `Value subclass:`: Association, DateTime, Regex, Random, StackFrame, CompiledMethod, FileHandle
- **New `Value.bt`** stdlib source file defining `Object subclass: Value` base class
- **Codegen fix**: Added "Value" to `known_value_roots` in `is_actor_class()` — without this, classes extending Value would silently compile as actors since Value has no .bt file and is absent from the cross-file superclass index
- **Test fix**: Updated `superclass_chain_for_integer` to include Value in the expected chain (Number → Value → Object → ProtoObject)

## Classification Rationale

Only map-backed classes (with `$beamtalk_class` key) were reclassified to Value. Native Erlang types (Symbol/atom, Pid/pid, Integer/int, String/binary, List/list, Tuple/tuple) remain as `Object subclass:` since they are not map-backed.

## Test Plan

- [x] All 1875 Rust unit tests pass
- [x] All 237 stdlib bootstrap tests pass
- [x] All 621 BUnit tests pass
- [x] All 2208 Erlang runtime tests pass
- [x] E2E tests pass

## Summary by CodeRabbit

* **New Features**
  * Introduced a new Value base class for immutable, value-semantics types (documented usage and examples).

* **Refactor**
  * Standard library class hierarchy updated: many core classes (Association, Boolean, Collection, CompiledMethod, DateTime, Exception, FileHandle, Number, Random, Regex, StackFrame, and others) now inherit from Value, shifting them to value-type semantics.
  * Runtime mappings and tests adjusted to reflect the new hierarchy.


jamesc:
Addressed the nitpick from the CodeRabbit review: updated the stale comment in `is_actor_class` (mod.rs:1231-1235) to remove the now-incorrect claim that `Value` has no `.bt` file. The comment now describes the actual reason — incomplete cross-file superclass indexing during independent file compilation. Commit: 4b3eddad.

PR #1001, opened 2026-02-28
Fix flaky workspace integration test: stale port file causes auth failure on restart (BT-967)

jamesc:
## Summary

- Fixes flaky `test_get_or_start_workspace_lifecycle_integration` (~1/9 failure rate)
- Root cause: stale `port` file from a previous failed run causes port discovery loop to connect to wrong BEAM node (different cookie) → 30s auth failure loop
- Extracts `remove_stale_runtime_files()` helper to unify stale-file cleanup with consistent error propagation

## Changes

- `process.rs`: replace inline pid/port cleanup with `remove_stale_runtime_files()` call before spawning
- `storage.rs`: extract `remove_file_if_exists()` and `remove_stale_runtime_files()`; `cleanup_stale_node_info` delegates to it; fixes silent `let _ =` error swallowing on port/pid cleanup

## Test plan
- [x] All 9 workspace integration tests pass
- [x] `just build && just clippy && just fmt-check` clean

Linear: https://linear.app/beamtalk/issue/BT-967

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cleanup of stale runtime artifacts (PID/port/node.info) when starting detached processes to avoid startup conflicts and improve reliability.
* **Tests**
  * Added a regression integration test ensuring stale port files are ignored and new nodes start cleanly.

### DIFF
--- a/crates/beamtalk-cli/src/commands/workspace/mod.rs
+++ b/crates/beamtalk-cli/src/commands/workspace/mod.rs
@@ -1224,6 +1224,11 @@ mod tests {
         .expect("first start should succeed");
         let stale_port = first_info.port;
         kill_node_raw(&first_info);
+        // Wait for epmd to deregister the old node name before restarting.
+        // After force-kill, epmd may still hold the registration briefly,
+        // which prevents a new node with the same name from starting.
+        wait_for_epmd_deregistration(&first_info.node_name, 5)
+            .expect("epmd should deregister node name within timeout");
 
         // Step 2: Remove node.info but deliberately LEAVE the stale port file.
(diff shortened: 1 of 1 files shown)


jamesc:
Addressed CodeRabbit review: added `wait_for_epmd_deregistration(&first_info.node_name, 5)` after `kill_node_raw` in the `test_start_detached_node_ignores_stale_port_file_integration` regression test (commit 1fc9a6c1). This matches the existing pattern at line 1149 and prevents an intermittent test failure from epmd name collision during immediate node restart.

PR #2051, opened 2026-04-21
Generic unification fails on nested, union, and FFI type shapes (BT-2023)

jamesc:
## Summary

Fixes three edge cases where generic type parameter unification fails, causing silent fallback to Dynamic:

- **A. Nullable unions**: `List(String) | Nil` arguments now unify against `List(T)` params by extracting the non-Nil Known member. Covers `collect:`, `select:`, `do:`, `detect:` on nullable collections.
- **B. Nested generics**: `resolve_type_param` now recurses into parametric forms like `List(E)` and `Dictionary(K, List(V))`, enabling block params typed as e.g. `Block(List(E), List(R))` to resolve correctly.
- **C. FFI polymorphic returns**: Call-site type args flow through FFI return types when param and return share the same base class (e.g., `lists:reverse(List(String))` returns `List(String)` instead of bare `List`).

All three bugs share the same root cause: the unifier assumed `Known(class, args)` and bailed on anything else.

## Key Changes

- `extract_non_nil_known`: new helper to unwrap nullable unions for unification
- `infer_method_local_params`: uses effective_arg from union unwrapping for both plain and parametric type params
- `resolve_type_param`: recursive handling of nested generic type strings
- `substitute_ffi_return_type`: new method to flow call-site type args through FFI returns
- 14 new regression tests covering all three sub-bugs plus edge cases

## Test Plan

- [x] All 3082 beamtalk-core tests pass (14 new)
- [x] Clippy clean (warnings = errors)
- [x] Format check passes
- [x] Multi-member unions correctly refuse to unify (ambiguous case)
- [x] Bare type params and known classes still resolve correctly (regression)
- [x] FFI returns with existing type args are not overridden

Linear: https://linear.app/beamtalk/issue/BT-2023

https://claude.ai/code/session_01SfCmrkrALjxnUeSbunSsDy

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved type inference accuracy for external function interface calls
  * Enhanced resolution of complex generic types with nested parameters
  * Fixed type parameter handling in nullable union scenarios

* **Tests**
  * Added comprehensive test coverage for generic type resolution and external function interface type handling

### DIFF
--- a/crates/beamtalk-core/src/semantic_analysis/type_checker/inference.rs
+++ b/crates/beamtalk-core/src/semantic_analysis/type_checker/inference.rs
@@ -1176,7 +1176,76 @@ impl TypeChecker {
         // Check argument types positionally against declared params
         self.check_ffi_argument_types(module_name, &function_name, &sig, arg_types, span);
 
-        sig.return_type
+        // BT-2023(C): Thread call-site type args through FFI return types.
+        // When the return type is a bare Known type (no type args) and a param
+        // shares the same base class, flow the call-site argument's type args
+        // to the return. E.g., `lists:reverse(List(String))` → `List(String)`.
+        Self::substitute_ffi_return_type(&sig, arg_types)
+    }
+
+    /// BT-2023(C): Substitute FFI return type with call-site type args.
+    ///
+    /// When an FFI signature has a bare return type (e.g., `List`) that matches
+    /// a param type, and the call-site argument carries type args (e.g.,
+    /// `List(String)`), flow those type args to the return type.
+    ///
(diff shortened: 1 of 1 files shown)


jamesc:
Closing this PR — it was created by an agent in a Wave 1 run that was aborted when the epic (BT-2024) execution plan was restructured. The new plan lands a shared framework first (BT-2025: centralised parametric type resolver) so the per-site fixes here can reuse it instead of introducing parallel ad-hoc helpers.

The tests added in this PR (14 regression cases for nullable unions, nested generics, and FFI polymorphic returns) are valuable and will be carried forward into the redo on top of BT-2025.

No action needed — will reopen as a new PR against BT-2025 once the framework lands.

---

PR #2053, opened 2026-04-21
fix: resolve Nil to UndefinedObject so isNil narrowing propagates (BT-2016)

jamesc:
## Summary

- `resolve_type_keyword` only mapped lowercase `nil` to `UndefinedObject`, but users write `Integer | Nil` (capital N) in type annotations. The parser preserves `Nil` as `Known("Nil")`, which `non_nil_type` didn't recognize, causing early-return narrowing (`ms isNil ifTrue: [^nil]`) to silently fail.
- Added `"Nil"` to both `resolve_type_keyword` (inference) and `resolve_type_keyword_static` (validation) so `Nil` canonicalizes to `UndefinedObject` consistently.
- Added `"Nil"` checks to `non_nil_type` as a defense-in-depth measure for any `"Nil"` values that bypass `resolve_type_keyword`.

## Test plan

- [x] 3 end-to-end regression tests parsing real Beamtalk source: keyword arg, typed assignment, instance send arg
- [x] 2 unit tests for `non_nil_type` directly: strip capital `Nil`, strip both `UndefinedObject` and `Nil`
- [x] All 3077 existing tests pass (0 failures)
- [x] Clippy clean

Closes BT-2016

https://claude.ai/code/session_01SfCmrkrALjxnUeSbunSsDy

## Summary by CodeRabbit

* **Bug Fixes**
  * Consistently treat both `nil` and `Nil` as the same nil type, improving type resolution.
  * Non-nil type narrowing now correctly excludes nil members (including both representations), so narrowed types propagate through keyword args, local assignments, and method arguments.

* **Tests**
  * Added tests validating nil handling and type-narrowing behavior across the noted call/assignment scenarios.

### DIFF
--- a/crates/beamtalk-core/src/semantic_analysis/type_checker/inference.rs
+++ b/crates/beamtalk-core/src/semantic_analysis/type_checker/inference.rs
@@ -308,7 +308,10 @@ impl TypeChecker {
     /// - Everything else passes through unchanged.
     fn resolve_type_keyword(name: &EcoString) -> EcoString {
         match name.as_str() {
-            "nil" => "UndefinedObject".into(),
+            // BT-2016: Both `nil` (lowercase keyword) and `Nil` (class name) map
+            // to the canonical internal name `UndefinedObject` so narrowing,
+            // non_nil_type, and all downstream comparisons work consistently.
+            "nil" | "Nil" => "UndefinedObject".into(),
             "false" => "False".into(),
             "true" => "True".into(),
             _ => name.clone(),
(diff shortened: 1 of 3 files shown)


jamesc:
Addressed Copilot review feedback in 288d21d:

- **Test filter (both flagged E2E tests + the third for consistency):** switched `filter(|d| d.message.contains("Type mismatch"))` → `filter(|d| d.category == Some(DiagnosticCategory::Type))`. Confirmed argument-type warnings format as `"Argument ... expects ... got ..."` at `method_validators.rs:480+`, `validation.rs:553+`, and `protocol.rs:507+`, all with category `Type`.
- **Side effect caught by the tightened filter:** the instance-send test was declaring `Object subclass: Receiver` and then calling `Receiver new`. Object subclasses aren't instantiable, so the test always emitted an `"Object-kind class ... cannot be instantiated with `new`"` type-category error that the loose filter hid. Fixed by declaring `Value subclass: Receiver`. The test now exercises its actual claim — narrowing propagation into instance-send arg positions.
- **Doc comment:** `resolve_type_keyword` doc now mentions `Nil` → `UndefinedObject` alongside the lowercase form, with a BT-2016 reference explaining why both spellings canonicalize.
- **Bonus (required to pass clippy):** dropped unnecessary `#` from three `r#"..."#` raw strings that contain no `"` — `needless_raw_string_hashes`.

All 3 narrowing tests + 540 other type_checker tests still pass; clippy + fmt clean.

---

PR #2082, opened 2026-04-23
Type ifNil:ifNotNil: expression as branch-union (BT-2047)

jamesc:
## Summary

Resolves [BT-2047](https://linear.app/beamtalk/issue/BT-2047). The overall `receiver ifNil: [a] ifNotNil: [:x | b]` expression now types as `typeof(a) | typeof(b)` (deduplicated) instead of `Dynamic`. BT-2046 already inferred both branches as `Block(..., R)` via `infer_args_for_if_not_nil` / `infer_if_not_nil_block`; this change consumes the returned type args at the send site to construct the union.

A branch whose body contains a non-local return (`^`) exits the enclosing method before its value is observed, so that branch contributes `Never` to the union — `union_of` skips `Never`, leaving the surviving branch's type. Applies to both keyword orders (`ifNil:ifNotNil:` and `ifNotNil:ifNil:`).

## Key changes

- `crates/beamtalk-core/src/semantic_analysis/type_checker/inference.rs`:
  - Short-circuit the send's return-type computation in `infer_message_send_with_receiver_ty` for `ifNil:ifNotNil:` / `ifNotNil:ifNil:` when both arg types are well-formed `Block(...)`.
  - New helper `if_nil_branch_union_ret_ty` extracts each branch's return type (last type-arg of its `Block`) and unions them; blocks containing `^` contribute `Never`.
- `crates/beamtalk-core/src/semantic_analysis/type_checker/tests.rs`: 5 new tests covering both keyword orders, dedup when both branches agree, and divergence (with matching and distinct branch types).

## Acceptance criteria

- [x] `ifNil: [a] ifNotNil: [:x | b]` has inferred return type `typeof(a) | typeof(b)` (deduplicated)
- [x] Same for `ifNotNil:ifNil:` (reversed order)
- [x] When one branch returns (`^...`), the expression's type comes from the other branch
- [x] Test added in type-checker test suite with the minimal repro from the issue
- [ ] `@expect type` at `beamtalk-exdura/src/workflow_runner.bt:71` (external repo, not in this worktree)

## Test plan

- [x] `cargo test -p beamtalk-core --lib bt2047` — 5 new tests pass
- [x] `cargo test -p beamtalk-core --lib` — all 3201 lib tests pass, no regressions
- [x] `just test` — Rust (3201) + stdlib (250) + BUnit (1899/1901) + runtime (3485 + 1242) pass
- [x] `just clippy` + `just fmt` clean

## Summary by CodeRabbit

* **New Features**
  * Improved return-type inference for conditional message sends (`ifNil:ifNotNil:` and `ifNotNil:ifNil:`): the result is the union of branch return types, with branches that perform non-local returns excluded from the union so diverging arms don't widen inferred types.
* **Tests**
  * New tests verifying union inference, deduplication when branches match, and correct handling of diverging (non-local-returning) branches.

### DIFF
--- a/crates/beamtalk-core/src/semantic_analysis/type_checker/inference.rs
+++ b/crates/beamtalk-core/src/semantic_analysis/type_checker/inference.rs
@@ -976,9 +976,12 @@ impl TypeChecker {
         // Restricted to non-class-side receivers: `ClassName ifNil: ... ifNotNil: ...`
         // and `self ifNil: ... ifNotNil: ...` inside a class method must
         // flow through `check_class_side_send` so an invalid metaclass send
-        // still emits DNU.
-        let is_class_side_receiver = matches!(receiver, Expression::ClassReference { .. })
-            || (env.in_class_method && Self::is_self_receiver(receiver));
+        // still emits DNU. Unwrap parens first so `(ClassName) ifNil: ...`
+        // and `(self) ifNil: ...` aren't accidentally treated as non-class-side.
+        let unwrapped_receiver = Self::unwrap_parens(receiver);
+        let is_class_side_receiver =
+            matches!(unwrapped_receiver, Expression::ClassReference { .. })
+                || (env.in_class_method && Self::is_self_receiver(unwrapped_receiver));
         if !is_class_side_receiver
             && matches!(
                 selector_name.as_str(),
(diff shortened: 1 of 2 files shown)


jamesc:
Fix pushed in 7807507f — `is_class_side_receiver` now unwraps parens, so `(ClassName) ifNil: ...` and `(self) ifNil: ...` correctly flow through `check_class_side_send`.

PR #2084, opened 2026-04-23
Block param inferred Dynamic in typed class when iterable's element type is Dynamic (BT-2042)

jamesc:
## Summary

Fixes the BT-2042 false-positive where iterating a Dynamic value inside a `typed` class forced users to annotate every block parameter.

When `blockers anySatisfy: [:b | ...]` ran against a Dynamic `blockers` (e.g., the result of `Dictionary at:ifAbsent:`), the block param `b` defaulted to `Dynamic(UnannotatedParam)`, and every use of `b` inside the body re-fired the "expression inferred as Dynamic in typed class" warning — pointing the user at a usage deep in the block body rather than the actual root cause (the Dynamic iterable).

The fix propagates `Dynamic(DynamicReceiver)` into block params in the fallback paths of `infer_args_with_block_context` when the receiver itself is Dynamic. `DynamicReceiver` is already filtered from the BT-1914 warning (matching how the send's own result is classified), so block-param usages no longer double-warn. Non-Dynamic fallback shapes (Union, Never, etc.) keep their existing behavior.

- Linear issue: https://linear.app/beamtalk/issue/BT-2042

## Changes

- `crates/beamtalk-core/src/semantic_analysis/type_checker/inference.rs`: added `infer_args_with_dynamic_block_params` helper and routed both early-bail paths (non-Known receiver, unresolved selector) through it.
- `crates/beamtalk-core/src/semantic_analysis/type_checker/tests.rs`: added three regression tests (repro from the issue, known-receiver regression guard, and annotated-local workaround).

## Test plan

- [x] `just clippy` clean
- [x] `just test` — all 3199 Rust lib tests pass (3 new), stdlib 250/250, BUnit 1899/1899, runtime 4727/4727
- [x] Manual repro from the issue (`/tmp/bt-2042-repro/filter.bt`) now lints clean
- [x] Workaround forms (annotated local, `@expect type`) still lint clean
- [x] Known-receiver path (`List(Dictionary)` field) still resolves block param correctly

Acceptance criteria:
- [x] Repro does not emit the Dynamic warning
- [x] Regression test in type-system suite covering `typed class + iterate over Dictionary at:ifAbsent: result`
- [x] Existing workarounds (`@expect type`, local type annotation) continue to work

Note: the issue lists block-parameter annotation `[:b :: Dictionary | ...]` as a workaround, but the current parser does not accept that syntax; only the `@expect type` and local-annotation workarounds are real today.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved block-parameter type inference when calls involve dynamic receivers, eliminating spurious "expression inferred as Dynamic" diagnostics in typed classes.
* **Tests**
  * Added tests validating block-parameter typing for sends with dynamic and known collection receivers to prevent regressions.

### DIFF
--- a/crates/beamtalk-core/src/semantic_analysis/type_checker/inference.rs
+++ b/crates/beamtalk-core/src/semantic_analysis/type_checker/inference.rs
@@ -2646,21 +2646,27 @@ impl TypeChecker {
 
         arguments
             .iter()
-            .map(|arg| match arg {
-                Expression::Block(block) if propagate_dynamic => {
-                    let param_types: Vec<InferredType> = (0..block.parameters.len())
-                        .map(|_| InferredType::Dynamic(DynamicReason::DynamicReceiver))
-                        .collect();
-                    self.infer_block_with_typed_params(
-                        block,
-                        arg.span(),
-                        &param_types,
-                        hierarchy,
-                        env,
-                        in_abstract_method,
-                    )
+            .map(|arg| {
+                // Unwrap parens so `foo: ([:x | ...])` gets the same
+                // Dynamic(DynamicReceiver) propagation as the unparenthesised
+                // `foo: [:x | ...]` form.
+                let inner = Self::unwrap_parens(arg);
+                if let Expression::Block(block) = inner {
+                    if propagate_dynamic {
+                        let param_types: Vec<InferredType> = (0..block.parameters.len())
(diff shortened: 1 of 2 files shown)


jamesc:
Fix pushed in 716967ef — `infer_args_with_dynamic_block_params` now unwraps parens before matching on `Block`, so parenthesized block literals get the same `Dynamic(DynamicReceiver)` param propagation.

PR #2130, opened 2026-04-26
Fix wrong asString assertion and add missing Character method coverage (BT-2094)

jamesc:
## Summary

Fixes a confirmed wrong assertion in `character_test.bt` and adds coverage for five documented Character methods that had no tests.

**Linear:** https://linear.app/beamtalk/issue/BT-2094

## Changes

- **Fix** `$A asString equals: "65"` → `equals: "A"`. The `Character>>asString` primitive maps to `beamtalk_character:as_string/1` which returns `unicode:characters_to_binary([65])` = `<<"A">>`, not the integer decimal string. Confirmed by the EUnit test in `beamtalk_character_tests.erl` and the `Character.bt` docstring.

- **Add** `$\r isWhitespace` and `(Character value: 32) isWhitespace` cases to `testCharacterOperations` (space and carriage-return were the only common whitespace characters not covered).

- **Add** `testAsIntegerAndPrintString` — covers `asInteger` (identity primitive) and `printString` (`$A` → `"$A"` format).

- **Add** `testUppercaseLowercase` — covers `uppercase`/`lowercase` round-trips for lower→upper, upper→lower, and already-cased inputs.

- **Add** `testCharacterValue` — covers `Character value:` class method for valid codepoints (0, 65, max 0x10FFFF) and `#type_error` for invalid inputs (negative, surrogate 0xD800, overflow 0x110000).

## Test plan

- [ ] `just test-bunit stdlib/test/character_test.bt` passes
- [ ] `just test-bunit` passes (no regressions in other tests)

https://claude.ai/code/session_016Mvmm8Pm2BWk8Sf9KNocgQ

---
_Generated by [Claude Code](https://claude.ai/code/session_016Mvmm8Pm2BWk8Sf9KNocgQ)_

## Summary by CodeRabbit

* **Tests**
  * Improved character test coverage by explicitly validating that carriage return and the space character are treated as whitespace. This strengthens whitespace handling checks and reduces the risk of regressions in character classification across input parsing and formatting.

### DIFF
--- a/stdlib/test/character_test.bt
+++ b/stdlib/test/character_test.bt
@@ -31,8 +31,35 @@ TestCase subclass: CharacterTest
     self deny: $A isLowercase
     self assert: $\n isWhitespace
     self assert: $\t isWhitespace
+    self assert: $\r isWhitespace
+    self assert: (Character value: 32) isWhitespace
     self deny: $A isWhitespace
     self assert: ($A + 1) equals: 66
     self assert: ($z - $a) equals: 25
-    self assert: $A asString equals: "65"
+    self assert: $A asString equals: "A"
     self assert: $A asFloat equals: 65 asFloat
+
+  testAsIntegerAndPrintString =>
+    self assert: $A asInteger equals: 65
+    self assert: $a asInteger equals: 97
+    self assert: $0 asInteger equals: 48
+    self assert: $A printString equals: "$A"
+    self assert: $0 printString equals: "$0"
+    self assert: $z printString equals: "$z"
+
+  testUppercaseLowercase =>
+    self assert: $a uppercase equals: 65
+    self assert: $A uppercase equals: 65
+    self assert: $z uppercase equals: 90
+    self assert: $A lowercase equals: 97
+    self assert: $a lowercase equals: 97
+    self assert: $Z lowercase equals: 122
+
+  testCharacterValue =>
+    self assert: (Character value: 65) equals: 65
+    self assert: ((Character value: 65) =:= $A)
+    self assert: (Character value: 0) equals: 0
+    self assert: (Character value: 1114111) equals: 1114111
+    self should: [Character value: -1] raise: #type_error
+    self should: [Character value: 55296] raise: #type_error
+    self should: [Character value: 1114112] raise: #type_error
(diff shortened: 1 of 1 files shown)


jamesc:
We're these methods documented but not available? Or did you hallucinate them?

jamesc:
They're documented — `asInteger`, `uppercase`, `lowercase`, and `value:` are all defined in `Character.bt` as `@primitive` directives with codegen mappings in `character.rs`. So not hallucinated.

The problem was that the tests used dynamic dispatch (integer runtime), but these methods only work via static dispatch when the compiler knows the value is a `Character` type. In BUnit, the character literal `$A` is just an integer at runtime, and dispatch goes through Integer's method table — which doesn't include `asInteger` or `uppercase`. So calling them causes a DNU even though the methods exist.

The net result of this PR is minimal: just `$\r isWhitespace` and `32 isWhitespace` (space via codepoint) added to `testCharacterOperations`. Proper coverage of `asInteger`, `uppercase`/`lowercase`, and `Character value:` would need either static type annotations in the test so the compiler can inline the Character dispatch, or a separate mechanism to test Character-specific primitives directly.

---

jamesc:
Seems like we should have a path to test these, no? Otherwise how do they get used by developer?

jamesc:
Yes, this is a real gap. Local type annotations (`c :: Character := $A`) are erased at codegen with no runtime effect, so they don't help. The compiler should be statically specializing `$A asInteger` to Character dispatch since `$A` is a Character literal and the type is statically known — but it's not doing that. At runtime `$A` lands in Integer's method table, which doesn't have `asInteger`, `uppercase`, or `lowercase`.

Two possible fixes:
1. Compiler: treat character literal receivers as Character-typed and use Character-specific primitive dispatch (the right long-term fix)
2. Short-term: cover these via EUnit tests on `beamtalk_character.erl` directly (`beamtalk_character:as_string(65)` etc.)

I'll create a follow-up issue to track the compiler gap.

---

jamesc:
Yep, seems like a bug, no?

jamesc:
Yes, it's a bug. Created BT-2095 to track it: https://linear.app/beamtalk/issue/BT-2095/character-literal-dispatch-specialize-dollara

---

PR #2425, opened 2026-06-02
Exclude test-harness/build-tooling crates from Rust coverage

jamesc:
## What

The workspace-wide `cargo llvm-cov` run counted crates that cannot be meaningfully unit-tested, which deflated the product-code coverage number reported by the badge. This excludes them:

| Excluded | Why |
|---|---|
| `beamtalk-parity-tests` | cross-surface test harness (`publish = false`) |
| `beamtalk-build` | build-script helpers, run during `build.rs` |
| `build-corpus/src/main.rs` | one-shot corpus-generation tool |

The exclusion list lives in one place — a new `llvm_cov_ignore` variable in the `Justfile` — shared by `coverage-rust` (HTML) and a new `coverage-rust-cobertura` recipe. CI now calls that recipe instead of an inline `cargo llvm-cov` command, so the badge and local runs apply the identical filter.

## What is *not* excluded (deliberately)

- **`beamtalk-examples` library** — already ~99% covered; only its `build-corpus` generator binary drags it down, so just that file is filtered.
- **`beamtalk-surface-drift`** — 75.7% covered; excluding it would *lower* the number.

## Impact

Cobertura line-rate (the badge metric): **79.5% → 80.5%**, by removing ~1,900 lines of non-product code from the denominator. No product code is hidden.

## Verification

- `cargo llvm-cov report --ignore-filename-regex '<list>'` confirms the new totals.
- `just --show coverage-rust` / `coverage-rust-cobertura` confirm the recipes render the variable correctly.

## Summary by CodeRabbit

* **Chores**
  * Updated CI/CD workflow to invoke unified Rust coverage task execution
  * Added Cobertura XML output format support for coverage badge integration in CI pipelines
  * Enhanced coverage configuration with filename filtering to exclude test harnesses and build-related tooling from reported metrics


jamesc:
@coderabbitai review

PR #2429, opened 2026-06-02
Enable warnings_as_errors for Erlang src; convert deprecated catch

jamesc:
## What

Adds `warnings_as_errors` to the runtime `erl_opts`, so Erlang compile warnings fail the build — matching the policy Rust already enforces (`cargo clippy -- -D warnings`). Part of the "keep all languages clean" ratchet.

This gates the **src/** compile; the `test` profile keeps `[debug_info]` so test helpers aren't held to the same bar (empirically the full eunit suite already compiles clean under the flag anyway).

## The only blocker: OTP 27's deprecated `catch`

OTP 27 warns on the legacy `catch Expr` expression form, which was the **only** thing failing a clean warnings-as-errors build in our code. Converted all 7 sites to `try ... catch ... end`:

| Site | Conversion |
|---|---|
| `class_registry` (ets cleanup), `object_class` (metadata/ets/pg cleanup) | fire-and-forget → `try X catch _:_ -> ok end` |
| `repl_json` (`printString`, `inherits_from`), `compiler_server` (`__beamtalk_meta`) | value-using → faithful `try X of … catch _:_ -> <original fallback> end` |
| `compile_fixtures.escript` (`port_close` on timeout) | cleanup → `try … catch _:_ -> ok end` |

Each value-using conversion preserves the exact fallback the old `catch` produced when the call raised.

## Not affected

Dependency warnings (ranch's `catch`, cowboy's exported-variable) compile with **their own** `erl_opts`, so `warnings_as_errors` does not propagate to them.

## Verification

- Clean `src` compile under the flag → no failures.
- erlfmt clean on all changed files.
- Full `beamtalk_runtime,beamtalk_workspace,beamtalk_compiler` eunit: **4252 tests**, compiles + runs. The lone failure is the **pre-existing** TMPDIR-path test (`status_returns_ok_map_when_meta_started`), unrelated to this change.

Independent of the other open coverage PRs.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Refactor**
  * Improved error handling across runtime components for more robust class registration, object shutdown, JSON conversion, and REPL output—reducing failures during cleanup and lookup operations.
  * Made port timeout handling more resilient to unexpected errors.

* **Chores**
  * Compiler configuration updated to treat warnings as errors for stricter code quality.

PR #2430, opened 2026-06-02
Enforce warnings-as-errors on stdlib build; fix SystemNavigation warnings

jamesc:
## What

Brings the Beamtalk stdlib build under the same warnings-as-errors policy as Rust (`clippy -D warnings`) and Erlang src (#2429). Two parts:

### 1. Fix the existing warnings
`SystemNavigation.bt` had three sites where `entry at: 3` — the extension source, FFI-typed `String | Binary` but always source text — flowed into String-expecting operations (`matchesRegex:`, `appendAllSendsIn:`, `addSentSelectorsIn:`), producing union-member dnu / arg-type warnings. The **BT-2254 comments already documented the intent to suppress these** ("suppress the union-member dnu", "suppress the union arg-type warning"); the sanctioned `@expect type` annotation was just never applied. Added it at all three sites.

### 2. Enforce going forward
Added a `--warnings-as-errors` flag to the `build-stdlib` CLI command (plumbed into `CompilerOptions`, which the compiler already honours), and the `build-stdlib` Justfile recipe now passes it — so every `just build` fails on a new stdlib warning.

## Verification

- `just build-stdlib` builds all **81 modules clean** with enforcement on.
- End-to-end enforcement check: temporarily removing one `@expect` reintroduces the warning and fails the build (`× Failed to compile 'stdlib/src/SystemNavigation.bt'`) — confirming the flag is not a no-op.
- rustfmt + clippy clean on the CLI changes.

## Context

Completes the "warnings = errors across all three languages" goal:
| Language | Status |
|---|---|
| Rust | already enforced (`clippy -- -D warnings`) |
| Erlang | #2429 |
| **Beamtalk** | **this PR** |

## Release Notes

* **New Features**
  * Added `--warnings-as-errors` flag to stdlib build command for stricter compilation validation.

* **Improvements**
  * Enhanced stdlib type safety with additional type annotations.

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -139,7 +139,7 @@ build-erlang:
 # Build standard library (stdlib/src/*.bt → BEAM, incremental — skips if up to date)
 build-stdlib: build-rust build-erlang
     @echo "🔨 Building standard library..."
-    @cargo run --bin beamtalk --quiet -- build-stdlib --quiet
+    @cargo run --bin beamtalk --quiet -- build-stdlib --quiet --warnings-as-errors
     @echo "✅ Stdlib build complete"
 
 # Build all example programs (examples/**/*.bt → BEAM)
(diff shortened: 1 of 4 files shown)


jamesc:
@coderabbitai review

PR #2499, opened 2026-06-08
refactor: extract sealed_fn_name() helper to selector_mangler BT-2403

jamesc:
Centralise the __sealed_ function-name prefix that was inlined via format!() at five codegen call sites. Add sealed_fn_name() to selector_mangler alongside safe_class_method_fn_name, using a prefix-aware atom-length guard (hashes selectors where __sealed_ + selector would exceed 255 chars). Output is identical for all realistic selectors (<= 246 chars).

## Summary by CodeRabbit

* **New Features**
  * Added `sealed_fn_name` helper function for generating sealed method function names.

* **Refactor**
  * Updated sealed method naming across code generation to use the new helper function, ensuring consistent and constraint-aware name generation.


jamesc:
@coderabbitai full review

PR #2560, opened 2026-06-12
Fix flaky e2e save/popover tests: wait explicitly for save banner (BT-2529)

jamesc:
## Summary

Fixes the intermittent `e2e` failure tracked in [BT-2529](https://linear.app/beamtalk/issue/BT-2529): the Senders/Implementors popover test (and its sibling ⌘S test) occasionally fails with

```
Could not find element "#method-editor" [text: "Saved increment on Counter"]
```

### Root cause

The Ctrl+S save in these browser tests is a **server round-trip**: `WorkspaceLive` compiles `Counter >> increment` via `Facade.dispatch(:save, …)` *before* it assigns `save_result` and renders the "Saved increment on Counter" banner. Under parallel CI load that compile + re-render regularly outlasts phoenix_test_playwright's **2s default assertion poll**, so the banner lands late and `assert_has("#method-editor", …)` misses it. 14/15 tests pass in each failing run — a timing race in the test, not a product bug.

### Fix

Add an explicit `timeout: 10_000` to the `assert_has` banner check in both affected tests:

- `the Senders/Implementors popover lists sites and opens one (BT-2495)` — the test named in the issue.
- `the KeyboardShortcuts hook compiles a method on ⌘/Ctrl+S (BT-2485)` — the sibling test with the identical save→assert race; fixed proactively to prevent the same flake.

The assertion still returns the instant the text appears, so passing runs are not slowed — only a genuine failure now waits the full window. The test is **not** deleted or skipped (per acceptance criteria).

## Test plan

- [x] Pre-push lint gate green (clippy, rustfmt, dialyzer, erlfmt, beamtalk-fmt, generated-spec validation)
- [ ] **`liveview-e2e.yml` CI lane** — this remote container has Rust + Erlang but no Elixir/Mix, so the Playwright suite and the ≥10× reliability loop must run in CI rather than locally.

## Acceptance criteria

- [x] Root-caused the race (save round-trip vs default assertion poll)
- [x] Test reliably waits for the editor state via a longer `timeout:` on the `assert_has` — test not deleted/skipped
- [ ] Repeated-run verification deferred to the e2e CI lane (no Elixir toolchain in this environment)

Closes BT-2529.

https://claude.ai/code/session_01XAnQonAeMqzktPxnpzBYZq

---
_Generated by [Claude Code](https://claude.ai/code/session_01XAnQonAeMqzktPxnpzBYZq)_


jamesc:
@coderabbitai full review

PR #2579, opened 2026-06-13
ADR 0097: Desktop Attach Client — One Front Node per Workspace

jamesc:
## Summary

Adds **ADR 0097** (status: Proposed) — a desktop client for the LiveView IDE
(`editors/liveview` / `bt_attach`) that attaches to **already-running** Beamtalk
workspaces, plus the ADR index entry.

## The decision

A thin **Tauri (Rust) desktop shell** whose main process acts as a **connection
broker**: for each workspace the user attaches to, it spawns **one dedicated
`bt_attach` BEAM node** (`PORT=<free> bin/server <id>`, reusing the existing
boot-time global-env attach model), pins a loopback bind + the unauthenticated
cookie-only path, and points a window at it. Disconnect kills the child process.

- **Connects to live workspaces** discovered from `~/.beamtalk/workspaces/`;
  does **not** spawn the Rust toolchain.
- **One front node per workspace** → blast-radius + crash isolation, and the
  single-target attach client (`workspace.ex`) stays near-untouched (one-line
  sname seed at most). Chosen over a single Phoenix node with per-session node
  targeting, which would concentrate multiple workspaces' cookies in one process.

## Key changes

- `docs/ADR/0097-desktop-attach-client-node-per-workspace.md` (new)
- `docs/ADR/README.md` (index entry)

## Review notes (already applied in-branch)

This went through `/review-adr`; commit `14f29e1` corrects load-bearing factual
errors caught against the code:

- The front **self-assigns** a unique sname in `ensure_distributed/0` and ignores
  `RELEASE_NODE`; node naming corrected accordingly.
- A BEAM **can** hold per-peer cookies (`set_cookie/2`), so the justification was
  reframed from "impossible" to **blast-radius isolation**.
- The prod release **binds all interfaces** and runs the OIDC fail-closed path by
  default — so the broker must force loopback + reject OIDC config (a documented
  **local-only posture**), otherwise it would regress ADR 0091.
- Added the **no-shell coordinator / PWA** alternative as the 80/20 challenger;
  the load-bearing spike must justify the Tauri shell against it.

Commit `fd5528a` documents **remote / OIDC attach as an explicitly out-of-scope,
separate topology** (future ADR) — it reuses none of the broker; the desktop role
there collapses to a persistent webview following the server-side OIDC redirect
flow, with a system-browser handoff for IdP embedded-webview restrictions.

## Notes

- Docs-only change — no Rust/Erlang touched, so CI build/clippy/fmt are N/A.
- No Linear issue (standalone docs branch). `/plan-adr` can break this into
  implementation issues when the ADR settles.

https://claude.ai/code/session_01Y1iLi7a5akjfvBjw4Xbb81

---
_Generated by [Claude Code](https://claude.ai/code/session_01Y1iLi7a5akjfvBjw4Xbb81)_


jamesc:
Thanks — all three Suggestions addressed in `a56e54c`:

**1. Probe contradiction.** Resolved by making the probe explicitly **two-stage** in both the Decision and Implementation §1/§2: (a) poll the HTTP port (Phoenix up), then (b) hit a new `GET /readiness` endpoint that forces `connect/0` + one cheap RPC and returns 200 only on workspace reachability. The ADR now states this endpoint **doesn't exist today** and is a small, additive front change (router + thin controller) — no longer described as "one-line."

**2. Loopback override mechanism.** Named it: a `BT_ATTACH_BIND_IP` env read in `config/runtime.exs` (defaulting to today's all-interfaces value so remote deploys are unchanged), which the broker sets to `127.0.0.1`/`::1`. Called out that `runtime.exs:74` has no env hook today, so this is part of the front-side work, not free.

**3. `SECRET_KEY_BASE` permissions.** Added: stored `0600` under the (already `0700`) workspace dir, with the same "forged cookie ⇒ RCE-equivalent" failure-mode framing as the bind.

I also corrected the now-inaccurate "one-line `workspace.ex` touch" claims throughout — the front-side work is three small additive pieces (sname seed, `BT_ATTACH_BIND_IP` hook, `/readiness` endpoint); the attach client's RPC/eval core stays untouched.

**Nits:** `BT-XXX` is intentionally a placeholder until `/plan-adr` files the implementation issues (the ADR is still Proposed); will be resolved before Accepted. Added the `rpc/3` function anchor to the one bare line-only citation; the other `workspace.ex` references already pair a function name for longevity.

---

PR #2641, opened 2026-06-18
System Browser: read-only native backing Erlang source for `native:` classes (BT-2578)

jamesc:
## Summary

Surfaces the backing Erlang module of a `native:` class (ADR 0056) in the LiveView IDE's System Browser, **read-only**. A native class only has `self delegate` facade methods — the real logic lives in the backing gen_server module's `handle_call`/`handle_info` clauses — so this exposes that module's source, with a best-effort jump from a delegating method to its clause.

Implements [BT-2578](https://linear.app/beamtalk/issue/BT-2578). Extends the ADR 0096 browse data source (op 5) over the ADR 0091 authenticated facade.

## What's in it

**Backend (Erlang)** — `beamtalk_repl_ops_browse.erl`
- New `browse-native-source` op: resolves the backing `.erl` via `module_info(compile)`, returns `content` + `source_origin` + `editable` (`project ⇒ true`) + a best-effort `handle_call` clause line-map (`clauses` + `selected_clause`). Degrades to `null` content ("source not available") for `.beam`-only builds; structured error for a non-native class.
- `native` + `backing_module` added to `browse-class-definition` (from the facade's `__beamtalk_meta/0`).
- `native_delegate` added to `browse-method-source`, derived from the **compiler's own signal** — the facade exports `dispatch_<selector>` only for `self delegate` methods — not a body-text heuristic.
- Pure reflection, `:read` capability (Observer-safe).

**UI (Elixir / LiveView)** — `facade.ex`, `workspace.ex`, `workspace_live.ex`, `app.css`
- `:read` facade dispatch for `browse_native_source` (no raw `:rpc`).
- Class-definition tab badges native classes ("Erlang backend: `<module>`") with a "View Erlang source" toggle → read-only pane (source path, read-only/editable origin, clause list, "source not available" empty state).
- A `self delegate` method tab shows "Native delegate → Erlang implementation"; clicking opens the def tab's native pane with the selector's `handle_call` clause highlighted, or explains where it completes (`handle_info`/helper) when there's no direct clause.

**Docs** — `surface-parity.md` documents `browse-native-source` as System Browser op 5.

## Tests
- Erlang: browse ops EUnit (clause parsing incl. comment/assignment anchoring, native fields, not-native error, `delegate_exported`). Full workspace EUnit **2192 green**.
- LiveView: `workspace_native_pane_test` — badge, lazy toggle, empty state, observer role, method→clause jump + highlight, no-matching-clause note, and full tree→definition→pane **navigation**. Bare `mix test` lane **202 green**.
- dialyzer, erlfmt, Elixir format + lint all green.

## Scope / follow-ups (not in this PR)
- **Read/write editing** of project-owned `native/` `.erl` (ADR 0082 loop) — explicitly out of scope.
- **Shared clause-mapping** between this op and the Rust `definition_provider` / BT-1215 — tracked as **BT-2582**.
- **Real-browser Playwright e2e** (`@moduletag :workspace`, needs a live node) — logic covered by the bare-lane LiveViewTest.
- `module_info(compile)` paths won't resolve in a stripped release → correct "source not available" empty state.

https://claude.ai/code/session_01Y2NwVstvMyKTfW7FWgXNdN

---
_Generated by [Claude Code](https://claude.ai/code/session_01Y2NwVstvMyKTfW7FWgXNdN)_


jamesc:
Thanks for the review. Triage of the latest round:

- **Nit (review attributions in comments)** — fixed in `27a5ff4`'s follow-up; the `(Claude review)`/`(CodeRabbit)` notes are removed from the source.
- **Suggestion 1 (`source_origin_of(Backing, BackingFile)` instead of `ModName`)** — declining: this would *regress* the common case. `beamtalk_class_registry:is_stdlib_module/1` matches **only** the `bt@stdlib@…` facade module-name pattern, so passing the plain backing atom (`beamtalk_subprocess`) returns `false`, falls through to `classify_source_origin/1`, and — since the stdlib `.erl` lives outside the user's project root — would classify `Subprocess` as `dependency` rather than `stdlib`. The facade `ModName` is the correct argument for the stdlib check. The edge case you describe (project facade wrapping a stdlib backing gen_server) is real but would need `is_stdlib_module/1` to also recognise backing modules; I've left it for the R/W follow-up rather than break stdlib detection now.
- **Suggestion 2 (two sequential RPCs in `browser_jump_native`)** — agreed it's a latency nit; both are `:read` ops and correctness is fine. Deferring as you suggest until the browse ops get batching.

All required CI checks are green on the latest HEAD.

---

PR #2683, opened 2026-06-21
feat(types): narrow singleton symbol types on (in)equality (BT-2617)

jamesc:
## Summary

Singleton symbol types (`#infinity`, `#ok`, …) already exist as first-class types (ADR 0068), but flow narrowing didn't account for equality tests against a singleton literal. A parameter typed `ms :: Integer | #infinity` stayed widened in *both* branches of `ms = #infinity ifTrue: [...] ifFalse: [...]`, which made precise singleton annotations awkward to consume.

This adds singleton-equality narrowing:

```beamtalk
withTimeout: ms :: Integer | #infinity -> TimeoutProxy =>
  ms = #infinity
    ifTrue:  [ "ms : #infinity here" ]
    ifFalse: [ "ms : Integer here" ]   // now narrows, previously Integer | #infinity
```

## Approach

A new `singleton_eq` narrowing rule (mirrors the existing rule-table pattern) detects `x = #foo` / `#foo = x` and the inequalities `/=` / `=/=`. Detection only sees the AST, so it records the tested singleton and a `negated` flag; `refine_singleton_narrowing` then resolves the variable's current type and splits it:

- **true branch** → the singleton (`#infinity`)
- **false branch** → the variable's type with that singleton removed (`Integer | #infinity` minus `#infinity` ⇒ `Integer`)
- **inequality** (`/=`, `=/=`) swaps the two branches

A `union_without` helper (the singleton counterpart of the existing `non_nil_type`) does the subtraction: a union collapses to its single remaining member, or to `Never` when the singleton was the only one left (the complementary branch is then unreachable).

This follows the precedent of `respondsTo:` / Result narrowing, where `detect` records intent and an apply-time refinement uses the variable's env type that the detector can't see.

## Also

Corrects a stale comment in the narrowing tests claiming union annotations "resolve to Dynamic" — they resolve to `InferredType::Union` via `type_resolver` (since BT-2016).

## Tests

- **Detection**: equality, symmetric `#foo = x`, inequality (`/=`, `=/=`), two-literal non-match (`#a = #b`)
- **Refinement**: union split, inequality branch-swap, multi-member union (`#north | #south | #east`), all-removed → `Never`
- 4152 `beamtalk-core` lib tests + 250 stdlib tests pass; clippy clean; `just build` + `just test-stdlib` green (no new warnings)

Closes BT-2617. Unblocks BT-2618 (tightening over-broad `Symbol` stdlib annotations).

---
_Generated by [Claude Code](https://claude.ai/code/session_01TQB3c9h5ii7TUYdoKskM7w)_

### DIFF
--- a/crates/beamtalk-core/src/semantic_analysis/type_checker/inference.rs
+++ b/crates/beamtalk-core/src/semantic_analysis/type_checker/inference.rs
@@ -2136,6 +2136,12 @@ impl TypeChecker {
                 members,
                 provenance,
             } => {
+                debug_assert!(
+                    members
+                        .iter()
+                        .all(|m| !matches!(m, InferredType::Union { .. })),
+                    "union members are kept flat by `union_of`, so a single subtraction pass suffices"
+                );
                 let kept: Vec<InferredType> = members
                     .iter()
                     .filter(|m| m.as_known().is_none_or(|n| n != removed))
@@ -2440,7 +2446,8 @@ impl TypeChecker {
                 // Single argument: narrow in the false branch (complement)
                 if let Some(arg) = arguments.first() {
                     if let Some(ref false_ty) = info.false_type {
-                        // Explicit false type (e.g., Result isOk/isError — BT-1859)
+                        // Explicit false type (e.g., Result isOk/isError — BT-1859;
+                        // or singleton (in)equality complement — BT-2617)
                         let ty = self.infer_block_with_narrowing(
                             arg,
                             &info.variable,
@@ -2486,7 +2493,8 @@ impl TypeChecker {
                 }
                 if let Some(false_arg) = arguments.get(1) {
                     if let Some(ref false_ty) = info.false_type {
-                        // Explicit false type (e.g., Result isOk/isError — BT-1859)
+                        // Explicit false type (e.g., Result isOk/isError — BT-1859;
+                        // or singleton (in)equality complement — BT-2617)
                         let ty = self.infer_block_with_narrowing(
                             false_arg,
                             &info.variable,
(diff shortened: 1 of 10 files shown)


jamesc:
Thanks for the review — addressed the nits in `6c2ea88`:

- **Stale `false_type` comment** (both `ifFalse:` and `ifTrue:ifFalse:` branches): now notes singleton narrowing as a second consumer of the explicit-`false_type` path.
- **`union_without` flat-union invariant**: added a `debug_assert!` that no member is itself a `Union` (kept flat by `union_of`).
- **RULES table comment**: now documents `singleton_eq`'s shape discriminant (binary op + exactly one Symbol-literal operand).

On the two larger suggestions, both turn out to be facets of how **singleton members inside a union** are type-checked — orthogonal to the narrowing itself — so I've tracked them in **BT-2624** rather than expand this PR:

- *Overly broad detection* (`x :: String` tested `x = #infinity` narrows silently): item 1.
- *End-to-end module test*: I investigated this and it's currently **not observable**. Singleton members in a union are handled conservatively by the union-send checker — `(Integer | #infinity) + 1` doesn't warn, and the guard itself either warns (`ms = #infinity` → *"Integer does not understand '='"*, since `Integer` has `=:=`/`=/=`/`/=` but not `=`) or infers `Dynamic` (`ms =:= #infinity`), which poisons the `ifTrue:ifFalse:` result. So a `check_module`-level test can't distinguish narrowed vs un-narrowed today. BT-2624 items 2–3 cover closing that gap, after which the e2e test becomes writable (also tracked there).

The detect + refine logic is covered by the 8 new unit tests; the apply glue is the shared narrowing path already exercised by the nil/Result rule tests.

---

PR #2755, opened 2026-06-24
Cache git repo toplevel per project path in beamtalk_git (BT-2621)

jamesc:
## Summary

BT-2608 made `beamtalk_git:run_git/2` resolve the git repository toplevel (`git rev-parse --show-toplevel`) from the project dir before **every** git op, so each op spawned two subprocesses. On the hot read path (`git_status` on every IDE git-panel refresh, `git_diff`) this doubled the subprocess round-trips.

The toplevel is static for a given project path, so this resolves it once and caches it.

## What changed

**`beamtalk_git.erl`** — `run_git/2` now calls a new `cached_repo_toplevel/2` instead of `repo_toplevel/2` directly. On a cache hit it skips the `rev-parse` subprocess; on a miss it resolves once and caches the result. Error results (not-a-repo, git absent) are deliberately left uncached so a later `git init` is picked up without a stale negative entry.

**`beamtalk_workspace_meta.erl`** — added `get_git_toplevel/1` (ETS read) and `set_git_toplevel/2` (cast). The cache is a single ETS row `{git_toplevel, ProjectPath, Toplevel}`:
- Stored as its own small row (**not** in `#state{}`), so the hot read path copies only a 3-tuple instead of the whole meta state (which carries class sources and loaded modules).
- Read straight from ETS — no gen_server round-trip on `git_status`/`git_diff`.
- Keyed by project path → self-invalidates on project switch (a lookup for a different path misses and forces re-resolution).
- Derived state — never written to `metadata.json`.

## Acceptance criteria (BT-2621)

- [x] `beamtalk_git` resolves the repo toplevel at most once per project path (cached).
- [x] Cache is invalidated/recomputed when `project_path` changes.
- [x] `git rev-parse` is not re-spawned on every `git_status`/`git_diff`/mutating op.
- [x] No behavioural regression vs BT-2608 (subdir-project revert/stage/unstage still correct).

## Testing

- Full `beamtalk_workspace` eunit: **2380 tests, 0 failures**, including 5 new tests: cache round-trip, project-path invalidation, no-disk-persistence, not-started graceful handling, and an end-to-end caching test through a real subdirectory git repo.
- `just build`, `rebar3 fmt --check`, and `just dialyzer` (no warnings) all pass.

---
_Generated by [Claude Code](https://claude.ai/code/session_01YTBGFZh28WjGduDoqgenJa)_

### DIFF
--- a/runtime/apps/beamtalk_workspace/src/beamtalk_git.erl
+++ b/runtime/apps/beamtalk_workspace/src/beamtalk_git.erl
@@ -252,12 +252,17 @@ mutate(Selector, Args, FailMsg) ->
 %% coincide; when it is a subdirectory (e.g. `examples/getting-started` inside
 %% the `beamtalk` repo) this prevents the path from being re-prefixed and
 %% matching nothing (BT-2608).
+%%
+%% The toplevel is static for a given project path, so it is resolved once and
+%% cached in beamtalk_workspace_meta (keyed by project path), removing the
+%% per-op `rev-parse` spawn on the hot read path (BT-2621). The cache
+%% self-invalidates when the project path changes (the keyed lookup misses).
 -spec run_git(atom(), [binary()]) ->
     {ok, binary(), non_neg_integer()} | {error, #beamtalk_error{}}.
 run_git(Selector, Args) ->
     case project_dir() of
         {ok, Dir} ->
-            case repo_toplevel(Selector, Dir) of
+            case cached_repo_toplevel(Selector, Dir) of
                 {ok, Toplevel} ->
                     run_git_in(Selector, Args, Toplevel);
                 {error, _} = Err ->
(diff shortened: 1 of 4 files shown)


jamesc:
Thanks for the review! Both nits are deliberate — each mirrors an established sibling pattern in the same file, so I've kept them for local consistency rather than introducing a one-off divergence:

- **`ets:whereis` guard in `handle_cast({set_git_toplevel, ...})`** — matches `store_state_in_ets/1` in the same module, which guards its `ets:insert` identically. Dropping the guard only in the new handler would make the two ETS-write sites in `beamtalk_workspace_meta.erl` inconsistent.
- **`timer:sleep(10)` in `stop_meta_if_running/0`** — mirrors the existing `stop_if_running/0` helper in `beamtalk_workspace_meta_tests.erl`, which does the same `gen_server:stop/1` + sleep.

You're right that both are technically redundant in the normal path. If we'd prefer the module move away from the defensive ETS-insert style / the post-stop sleep, I'm happy to do that as a small follow-up that updates the precedents too — but kept this PR consistent with the surrounding code. Non-blocking either way.

---

PR #2769, opened 2026-06-26
feat(runtime): connected-mode Program exit: ends the session (BT-2688)

jamesc:
ADR 0099 §3 / **Phase 5** of epic BT-2684, follow-up to the now-merged #2768. Implements the **connected-mode** half of `Program exit:` (the part Phase 3 gated with `#unsupported`).

In a **connected session** (REPL / MCP / LSP), `Program exit: N` now ends only *that session's* job and reports the status to the connecting client, instead of raising `#unsupported`. Run-mode / escript still halt the node via `erlang:halt/1` (unchanged).

## What changed

- **`beamtalk_program`** — the connected branch of `exit:` raises a tagged `{beamtalk_script_exit, N}` signal instead of `raise_connected_exit_unsupported/1` (gate removed).
- **`beamtalk_repl_eval`** — the eval worker catches the signal *ahead of* the generic handler → `{script_exit, N, …}` (a success shape, not an error).
- **`beamtalk_repl_shell`** — replies with the exit status, then stops the session shell (a `temporary` child of `beamtalk_session_sup`), leaving the node and every other session running.
- **Protocol** — a new `{script_exit, …}` op_result + `encode_script_exit/4` emit a `done` reply carrying a dedicated **`exit_code`** field (`value: null`); the Rust `ReplResponse` gains `exit_code: Option<u8>`.
- Range/type validation (`Program exit: 999` → `#invalid_argument`, `Program exit: "x"` → `#type_error`) and `System halt:` are unchanged.

## Testing

- **EUnit:** `beamtalk_program` (signal raised in connected mode + validation), `beamtalk_repl_shell` (reply + session termination, node stays up), `beamtalk_repl_protocol` (`encode_script_exit` new/legacy), `beamtalk_repl_eval` (`inject_output`).
- **Rust:** `ReplResponse` deserializes `exit_code`; absent on ordinary replies.
- `program_test.bt`'s old `#unsupported` expectation removed — the connected-exit behaviour now needs a live session shell, so it is covered in EUnit (a BUnit block has no session evaluator to catch the signal).
- Local: full `test-runtime` (5133 + 1541), BUnit (2608), Rust protocol tests, `clippy`, `dialyzer` — all green.

## Known limitation (documented in `beamtalk_program`)

A connected `Program exit:` wrapped in a *matching* `on:do:` (`on: Error do:` / `on: Exception do:` / catch-all) is intercepted as an `erlang_throw`. `ensure:` is unaffected (re-raises after cleanup), as are the common top-level / REPL paths. The robust fix is adding this signal to the `{'$bt_nlr', _}` non-local-return passthrough the `on:do:` codegen already emits — tracked as a follow-up.

## Follow-ups
- **BT-2690** — UAT test in `beamtalk-uat`.
- **BT-2691** — connected-mode `beamtalk run` script dispatch (the CLI consumer that turns `exit_code` into a process exit; `needs-spec`).

---
_Generated by [Claude Code](https://claude.ai/code/session_01XFJR7WPxr75jwuT5vjyW5c)_

### DIFF
--- a/crates/beamtalk-repl-protocol/src/response.rs
+++ b/crates/beamtalk-repl-protocol/src/response.rs
@@ -48,6 +48,13 @@ pub struct ReplResponse {
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub status: Option<Vec<String>>,
 
+    /// POSIX exit status from a connected-session `Program exit:` (BT-2688,
+    /// ADR 0099 §3). Present only on the `done` reply that ended the session; the
+    /// connecting client adopts it as its own process exit code. `None` for every
+    /// other response.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub exit_code: Option<u8>,
+
     // --- Common payload ---
     /// Result value (eval, clear, kill, clone, etc.).
     #[serde(default, skip_serializing_if = "Option::is_none")]
@@ -307,6 +314,23 @@ mod tests {
         let resp: ReplResponse = serde_json::from_str(json).unwrap();
         assert!(resp.is_error());
         assert_eq!(resp.error_message(), Some("Something went wrong"));
+    }
+
+    #[test]
+    fn deserialize_script_exit_response() {
+        // BT-2688: connected-session `Program exit: 5` — a `done` reply (not an
+        // error) carrying the POSIX exit status the CLI adopts.
+        let json = r#"{"id":"msg-008","value":null,"status":["done"],"exit_code":5}"#;
+        let resp: ReplResponse = serde_json::from_str(json).unwrap();
+        assert!(!resp.is_error());
+        assert_eq!(resp.exit_code, Some(5));
+    }
+
+    #[test]
+    fn exit_code_absent_on_ordinary_response() {
+        let json = r#"{"id":"msg-009","value":42,"status":["done"]}"#;
+        let resp: ReplResponse = serde_json::from_str(json).unwrap();
+        assert_eq!(resp.exit_code, None);
     }
 
     #[test]
--- a/crates/beamtalk-repl-protocol/src/response.rs
+++ b/crates/beamtalk-repl-protocol/src/response.rs
@@ -49,9 +49,11 @@ pub struct ReplResponse {
     pub status: Option<Vec<String>>,
 
     /// POSIX exit status from a connected-session `Program exit:` (BT-2688,
-    /// ADR 0099 §3). Present only on the `done` reply that ended the session; the
-    /// connecting client adopts it as its own process exit code. `None` for every
-    /// other response.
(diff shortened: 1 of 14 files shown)


jamesc:
Thanks for the review — all three suggestions addressed in `3c4f8f2`:

**1. Trace-mode `Program exit:`** — One correction: there was no actual `case_clause`. `do_eval_trace`'s generic `catch Class:Reason:Stacktrace` already wraps the `{beamtalk_script_exit, _}` throw as an `{error, …}`, so `handle_info` matched its existing `{error, …}` clause — no crash. But it *did* mean trace mode reported `Program exit:` as an error instead of exiting. Fixed properly: `do_eval_trace` now catches the signal ahead of the generic clause and returns `{script_exit, …}`, the `eval` op's trace branch passes it through, and the `do_eval_trace` / `eval_trace/2` specs carry the new shape — so a traced exit now ends the session consistently with the non-trace path.

**2. `eval_async` streaming subscriber** — Closed for surface parity: `beamtalk_ws_handler` now answers `{eval_script_exit, …}` by emitting the `encode_script_exit` reply (same `exit_code` JSON the synchronous op path produces) and clearing `pending_eval`, instead of leaving the subscriber without a terminal. The terminal-message set is now documented on `eval_async/3`.

**3. `exit_code` doc** — Reworded: it's the wire contract a connecting client *will* adopt once connected-mode `beamtalk run` dispatch lands (**BT-2691**); no consumer reads it yet, so it affects no CLI exit code today.

Local `build`, `dialyzer`, and `test-runtime` (5133 + 1549) green. CI re-running on the new commit.

---

jamesc:
Final polish in `b48ef89`:

**Suggestion 1 (LiveView `eval/2`)** — Confirmed it falls through to `{:unexpected_reply, …}`. Since the `exit_code` field doc already states no consumer reads it yet (BT-2691 owns cross-surface consumption), I added a doc breadcrumb on `eval/2` flagging the `{:script_exit, …}` term and pointing to BT-2691, rather than a half-arm that would just cascade to the UI caller. The session still terminates correctly today; this is the documented pre-BT-2691 state.

**Suggestion 2 (e2e REPL-protocol test)** — The `tests/repl-protocol` `.btscript` harness only matches `// =>` result values / `ERROR:` substrings; it can't assert the `exit_code` JSON field, and `Program exit:` would terminate the connection mid-file. So this isn't expressible there without a harness extension. The behaviour is covered by EUnit (shell + eval + protocol encoder) now, and the live-protocol e2e is exactly the UAT tracked in **BT-2690**.

**Nits** — `on:do:` caveat now cites BT-2688; `inject_output/3` spec narrowed from `tuple() -> tuple()` to the concrete 3-tuple → `eval_result()` union; the script_exit shell test kills its stand-in worker.

This is the final polish pass (build + dialyzer green). The review has passed with no blockers; merging on green.

---

PR #2777, opened 2026-06-27
stdlib: add aggregation + enumeration protocol to Collection (BT-2695)

jamesc:
## Summary

Lifts common aggregate and enumeration helpers onto the abstract `Collection` class so all six concrete collections (`List`, `Array`, `Set`, `Bag`, `Dictionary`, `Interval`) inherit them via `do:`/`inject:into:`. Closes the most visible gap from the stdlib audit — `#(1, 2, 3) sum` did not work before.

Implements **BT-2695** (part of the **BT-2692** "Expand stdlib" epic).

## What's added (on `Collection`)

| Method | Impl | Notes |
|---|---|---|
| `sum` | `@primitive` | `0` on empty |
| `max` / `min` | `@primitive` | raise `#beamtalk_error` on empty |
| `average` | `@primitive` | `Float`; raises on empty |
| `count:` | pure BT | inherited default; `List` keeps its `@primitive` override |
| `noneSatisfy:` | pure BT | `(anySatisfy:) not` |
| `do:separatedBy:` | pure BT | separator runs between elements only |
| `eachWithIndex:` | pure BT | lifted from `List`; redundant `List` copy removed |

`sum`/`max`/`min`/`average` are `@primitive` (backed by new `beamtalk_collection:sum/maximum/minimum/average`, built on the existing `to_list/1`) because a pure-BT fold performs arithmetic/comparison on the generic element type `E`, which the gradual type checker cannot prove is numeric.

## Files

- `stdlib/src/Collection.bt` — the new protocol
- `runtime/apps/beamtalk_stdlib/src/beamtalk_collection.erl` — aggregate helpers
- `crates/beamtalk-core/src/codegen/core_erlang/primitives/collection.rs` — lowering (+ unit test)
- `crates/.../generated_builtins.rs` — auto-regenerated method table
- `stdlib/src/List.bt` — drop redundant `eachWithIndex:`
- `stdlib/test/collection_protocol_test.bt` — aggregates on List/Set/Interval, empty-collection error cases, `eachWithIndex:`/`do:separatedBy:` behaviour
- `README.md` — fix a pre-existing `doc_markdown` clippy lint (`CodeMirror`) that blocked the lint gate

## Testing

- BUnit: 2620/2620 pass
- Rust unit tests, `just clippy`, `just dialyzer`: clean
- Empty-collection policy verified (`#() sum` → 0; `#() max`/`min`/`average` raise `#user_error`)

## Known limitation (follow-up: BT-2703)

Actor state-mutation inside `eachWithIndex:`/`do:separatedBy:` blocks is **not** supported — the state-threading codegen can't thread through these (especially the two-block `do:separatedBy:`), so it's left out of the selector registry. Non-mutating use works; mutation-in-block errors at compile time (not silently). Tracked in BT-2703.

`flatMap:`/`zip:` were intentionally **not** lifted — both are `@primitive` on `List` and order-sensitive (ill-defined for unordered `Set`/`Bag`).

---
_Generated by [Claude Code](https://claude.ai/code/session_01V8GmLeCfDFDVNqEy2dCcgW)_


jamesc:
Thanks for the review. Disposition of the findings (addressed in 75ebb2b):

**Accepted:**
- **#3 `noneSatisfy:` empty case** — added `self assert: (#() noneSatisfy: [:x | x > 0])`.
- **#5 moduledoc** — `beamtalk_collection` moduledoc now documents the four new aggregate exports.

**Accepted in spirit, but the suggested code can't be used — #4 (vacuous separator assert):** I tried your exact `count := count + 1` counter and it **compiles but fails at runtime**: `function expects 1 arguments but was called with 0` at `Collection>>do:separatedBy:`. Mutating a local inside the separator triggers the state-threading transform, which rewrites the block to take a threaded-state arg, but `do:separatedBy:` invokes the separator with `value` (0-arity) → arity mismatch. This is the BT-2703 limitation (now updated with this finding). So I dropped the misleading `[self assert: true]` for a no-op + comment; the conditional-firing guarantee is already covered by `testDoSeparatedBySingleElementHasNoSeparator` (separator = `[self assert: false]`, which fails if it ever fires for a singleton).

**Declined (intentional, non-blocking):**
- **#1 raw `badarith` on non-numeric elements** — kept. Consistent with how `inject:into:` and user blocks surface arith errors, and the runtime already wraps raw Erlang errors into a catchable `#beamtalk_error` at the exception boundary (`beamtalk_exception_handler`). The real improvement is enriching that central boundary with selector/class context rather than per-method `try/catch` — tracking separately.
- **#2 `average` two-pass** — kept. `lists:sum` + `length` are native loops; a hand-rolled `foldl` allocates a 2-tuple per element (no win, likely a loss vs two native passes), and `to_list` already dominates the cost. Clarity wins.

---

PR #2786, opened 2026-06-27
perf: leaner inject:into: via accumulator-last fold (BT-2713)

jamesc:
## Summary

`inject_into/3` wrapped the BT block in a `lists:foldl` arg-swapper. The BT-2692/2708 spike blamed that double-fun-indirection for the ~2.4× pure-BT fold tax — but that was **wrong**.

A controlled micro-benchmark shows the wrapper costs **~0%**: `lists:foldl`+wrapper and a hand-rolled fold with the accumulator in the *middle* arg both clock ~880 µs/op (~1.8× native). The real lever on the OTP 28 BEAM JIT is the loop's **accumulator-argument position**:

| fold loop shape (block still `Block(Acc, Elem)`) | µs/op | vs native |
|---|---|---|
| `lists:foldl` + arg-swap wrapper (old) | ~883 | ~1.78× |
| hand-rolled, accumulator in the *middle* `(List, Acc, Block)` | ~878 | ~1.77× |
| hand-rolled, accumulator as the *last* arg `(List, Block, Acc)` | ~516 | **~1.04×** |
| native `lists:foldl` baseline | ~497 | 1.00× |

Threading the accumulator as the loop's **last** argument lets the JIT keep it in a register the fun-call ABI doesn't disturb. `inject_into/3` now hand-rolls that accumulator-last fold, so **every** inject-based pure-BT fold is leaner at once — no change to the block call (BT blocks are unavoidably accumulator-first).

## Results

`sum`-via-`inject:` ratio: **~1.22× at N=100000** (was ~2.4×), ~0.99× at N=1000 — comfortably under the ≤1.5× target.

## Acceptance criteria

- [x] `sum`-via-`inject:` ratio meaningfully below 2.4× (target ≤ ~1.5×) → ~1.22×
- [x] All collection/inject tests pass unchanged; `just ci` clean
- [x] `benchmarks.md` updated with before/after + the JIT register-layout finding

## Notes

- Second commit regenerates `corpus.json`, which was stale on `main` (new getting-started examples added without rebuilding it). Unrelated to the fold change, but required for `just check-corpus` / CI to pass.


jamesc:
Thanks — addressing the review.

**Blockers:** none (review is PASS). ✅

**Suggestions** — both concern source files that live on `main` and are only *snapshotted* into `corpus.json` by the second commit; neither is modified by this PR (diff is `beamtalk_collection.erl` + `benchmarks.md` + `corpus.json`). Keeping them out of this perf PR to avoid scope creep:

1. **`ConsoleEcho.run` doc/code mismatch** — valid: the module doc says "Returns nil at EOF" but `run` ends with `self`. The `nil` describes `readLine:`s return, not the method. Real (cosmetic) doc bug in `examples/getting-started/src/console_echo.bt`; filing a follow-up rather than touching it here.
2. **`testMapDestructuringMissingKeyRaisesError` "not found" substring** — pre-existing test on `main` (BT-2707 reclassification). The redundancy-vs-`testMapDestructuringMissingKeyIsKeyError` point is a fair style call; noting for a follow-up, out of scope here.

**Nits:** `sum/1` keeping `lists:foldl` is intentional as you noted (native arithmetic, no BT block call — accumulator-last lever does not apply). `fold_block/3` comment style matches the codebase.

Production change verified locally: `just ci` clean, and a controlled micro-benchmark reproduces the documented table (wrapper-removal ~0%, accumulator-last ~516µs ≈ 1.04× native).

PR #2794, opened 2026-06-28
Actor codegen: strip __local__ threading temps from persisted state & watch notifications (BT-2717)

jamesc:
## Summary

Fixes [BT-2717](https://linear.app/beamtalk/issue/BT-2717). Actor methods that thread an outer local through a desugared control-flow construct (`eachWithIndex:`/`do:separatedBy:`, conditionals, counted loops) pack the local into the gen_server `State` under a `__local__<name>` key. These codegen-internal temporaries leaked into the **persisted actor state** and, for a **watched** actor, surfaced as spurious `__local__*` state-change notification slots.

The leak is systemic across multiple codegen emitters with no single emitter-level chokepoint, so the fix strips at the **outermost state-commit boundaries** (where the gen_server actually persists `NewState`), plus a runtime belt-and-braces for the notification path.

## Changes

**Codegen** (`gen_server/callbacks.rs`)
- Generated `handle_call` and `handle_cast` strip `__local__*` keys from the committed `NewState` via `beamtalk_actor:strip_local_temps/1` before notifying the per-object change substrate and replying (covers both the 3-tuple `PropCtx` and 2-tuple message clauses).
- `handle_continue` (post-initialize commit) does the same for `InitNewState`, so an `initialize` that threads a local doesn't persist a temp into the actor's first state.

**Runtime** (`beamtalk_actor.erl`)
- New `strip_local_temps/1` (+ `is_local_temp_key/1`): removes all `__local__`-prefixed atom keys; returns the map unchanged (no rebuild) when there are none — cheap on the common path.
- `changed_state_keys/2` also excludes `__local__`-prefixed keys (belt-and-braces for the notification path, regardless of leak source).

**Tests**
- EUnit: `strip_local_temps/1` unit coverage + an end-to-end watch-notification test driving the real `beamtalk_object_watch` substrate (a `__local__gaps` temp produces no changed slot while a real field still does).
- Codegen: extended the BT-2524 callback test and added a `handle_continue` strip test.
- Regenerated 62 actor codegen snapshots.

## Why this placement
Stripping only at the outermost gen_server callback is correct: stripping inside `safe_dispatch`/self-sends would corrupt an outer method's in-flight local threading (the inner call's returned state still carries the outer method's live `__local__` keys). Locals are always seeded fresh in-method before use, so stripping persisted state never affects re-execution.

## Verification
Local: workspace Rust tests (44 suites), stdlib (223), BUnit (2653), runtime EUnit (267), clippy, `cargo fmt`, `rebar3 fmt`, `rebar3 dialyzer` — all green.

## Follow-up
- **BT-2718** (filed from code review): reserve the internal `__…__` state-key namespace with a semantic validator so a user field/var named `__local__*` (or `__methods__`, etc.) is rejected rather than silently stripped. Latent today — no existing `.bt` uses such names.

---
_Generated by [Claude Code](https://claude.ai/code/session_011WUTHZTeNvGyRMKYcz5xZq)_

### DIFF
--- a/crates/beamtalk-core/src/codegen/core_erlang/gen_server/callbacks.rs
+++ b/crates/beamtalk-core/src/codegen/core_erlang/gen_server/callbacks.rs
@@ -1216,12 +1216,16 @@ impl CoreErlangGenerator {
                     nest(
                         INDENT,
                         docvec![
+                            // BT-2717: strip codegen-internal `__local__` threading
+                            // temporaries before persist + notify (see handle_call).
+                            line(),
+                            "let CleanCastNewState = call 'beamtalk_actor':'strip_local_temps'(CastNewState) in",
                             // BT-2524: same per-object change push as handle_call, for
                             // a state write committed via a fire-and-forget cast.
                             line(),
-                            "let _CastStateChanged = call 'beamtalk_actor':'notify_state_change'(State, CastNewState) in",
+                            "let _CastStateChanged = call 'beamtalk_actor':'notify_state_change'(State, CleanCastNewState) in",
                             line(),
-                            "{'noreply', CastNewState}",
+                            "{'noreply', CleanCastNewState}",
                         ]
                     ),
                     line(),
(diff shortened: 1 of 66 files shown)


jamesc:
Thanks — both addressed in `1d69d45`:

**Blocker — `handle_info` for Server subclasses:** Fixed. The Server-subclass `handle_info` success arm now strips the committed state via `let CleanInfoNewState = call 'beamtalk_actor':'strip_local_temps'(NewState) in` before `{'noreply', CleanInfoNewState}`, matching `handle_call`/`handle_cast`/`handle_continue`. `handle_info` doesn't call `notify_state_change`, so only the persist-side strip is applied. Added `test_bt2717_handle_info_strips_local_temps_for_server_subclass`, and the `server_handle_info.bt` BUnit fixture exercises it compiled-to-BEAM at runtime.

**Suggestion — stop arm in `generate_post_initialize_check`:** Changed `{'stop', Error, InitNewState}` → `InitCleanState` for consistency (terminating path, state discarded — no behavioural change). Validated by the `typed_no_default_actor.bt` BUnit fixture.

---

jamesc:
Thanks for the PASS. Triage of the non-blocking items:

**Error-path `ErrState` (suggestion):** No live gap in current codegen — `ErrState` is always the *clean incoming* `State`, never a threaded one. `safe_dispatch`'s catch returns `{'error', {Type, Error, Stacktrace}, State}` with the original `State` parameter (a method that fails mid-threading has its `let`-bound `StateAcc` mutations discarded on unwind), and the `dispatch/4` DNU/super error arms also return the incoming `State`. So no `__local__` key can reach the error arm to persist — consistent with the earlier review's correctness note. Not filing a follow-up since it isn't reachable; if a future codegen change ever returns a *threaded* state in an error tuple, the `changed_state_keys` belt-and-braces already shields notifications and this arm would be revisited then.

**Test nits:** Acknowledged. `bt2717_watch_teardown/1` deliberately mirrors the sibling `beamtalk_object_watch_tests.erl` convention (setup starts the server only if absent; teardown is a no-op — the server is normally part of the supervised tree in the runtime test app). The post-`subscribe/2` `sys:get_state` barrier is indeed redundant given the synchronous call — harmless; leaving it rather than re-spinning full CI for a test-only cosmetic. Will batch these into a future test-cleanup pass if we touch this file again.

---

PR #2818, opened 2026-07-02
Read-only pane + resolved docs for compiler-derived methods (BT-2714)

jamesc:
## Summary

Follow-up on BT-2714. Compiler-derived methods (value accessors, `with<Field>:` setters, actor `new`/`new:`/`spawn`/`spawn:`) were opened in the System Browser as a **blank editable buffer** — but they have no editable source. This makes them read-only and, crucially, surfaces their real documentation.

## Backend — resolve docs instead of nulling them

The `browse-method-source` op had its own logic that nulled `source`, `doc` **and** `signature` for a synthetic method. Only `source` needed to be null (a synthetic method has no editable body, and resolving it would surface an inherited implementation — e.g. `Actor`'s real `spawn` body — under the subclass's generated row). Doc/signature *should* flow through.

- New reusable `beamtalk_repl_docs:method_doc_signature_resolved/3` extracts the `{Doc, Signature}` resolution that `Beamtalk help:` already performs (walking the class hierarchy → `__doc__`/`__signature__`).
- `browse-method-source` now keeps `source: null` for a synthetic method but resolves its doc/signature via that helper — so a derived `spawn`/`new` shows `Actor`'s curated docs, a value accessor shows its signature, and the browser pane and `:help` can never disagree.

## LiveView — read-only derived pane

The method editor opens a synthetic method as a **read-only** tab: the doc block forced open in its normal position (no editor competes for the space) + a "compiler-derived, no editable source" note with the signature, instead of a blank CodeMirror buffer.

## Surface audit

- **MCP `docs`** already routes through `Beamtalk help:`, so it was always correct — no change.
- **VS Code / LSP hover** is a *separate* compile-time path (`HoverInfo`) that does not use the runtime resolver; surfacing synthetic docs there is out of scope (would need its own change).

## Testing

- `beamtalk_repl_docs:method_doc_signature_resolved/3` — 3 new EUnit tests (real curated doc reused, `{null,null}` fallback).
- LiveView — 2 new tests (derived method → read-only pane, no editable form; ordinary method → editor form).
- Green: Rust build · Dialyzer · erlfmt · Elixir format · workspace EUnit (2397/0) · docs EUnit (83/0) · LiveView (444/0). Full pre-push CI passed.

## Follow-ups

- **BT-2734** — generate `__doc__`/`__signature__` for value-type accessors at codegen (so value accessors show docs too, not just a signature).
- **BT-2735** — enrich the method-row hover with signature + doc.

🔗 https://linear.app/beamtalk/issue/BT-2714


jamesc:
Addressed both in e8de45fe:

- **"in image" note on synthetic tabs** — suppressed via `not synthetic_tab?(active_tab(assigns))` in the `:if` guard, with a new test asserting it's absent on the derived-method pane.
- **Stub `value` inconsistency** — added a clarifying comment on the `browse_method_source` `"value"` arm explaining it's intentionally modelled as `indexed` (real source body) for the existing badge-only tests, while `withValue:` is the consistent end-to-end synthetic example.

PR #2863, opened 2026-07-07
Epic BT-2748: Sendability Typing from Class Kinds (ADR 0103)

jamesc:
Implements ADR 0103 — derive a **sendability tier** for every inferred type from its class kind (ADR 0067) and warn when a value's tier is too weak for the boundary it crosses. Advisory only (ADR 0100): type-checker + one new keyword + a Phase 3 validator + hover — no codegen or runtime-dispatch change.

Executes epic **BT-2748** (6 child issues) sequentially on one branch, one commit each, in dependency order.

## What's in it

| Issue | Change |
|---|---|
| **BT-2753** | Tier lattice `Sendable < SendableRef < HandleScoped` + `Unknown`; shared `tier_of` core; builtin tier table (`Pid`→SendableRef, `Port`/`FileHandle`→HandleScoped(#process), `Reference`→Sendable, `Subscription`→HandleScoped(#node)); actor message-arg check; `SendableRef` hover; `DiagnosticCategory::Sendability` + `@expect sendability` |
| **BT-2754** | `handleScope: #symbol` class-side keyword — parser, `ClassDefinition` AST, `ClassInfo` field, inherited `ClassHierarchy::handle_scope` accessor, `__beamtalk_meta` emission (conditional), Value/Actor advisory |
| **BT-2755** | Full tier derivation — `Value` structural composition (weakest field), generic `type_args` (`List(Port)`→HandleScoped), user `handleScope:` consumption, `spawnWith:` map-value check, all-tier hover, stdlib inventory audit (documented in the ADR) |
| **BT-2756** | Block-capture Phase 3 validator joining `CapturedVar` × `type_map` — a block sent to another process (`!` cast, `Timer`, non-self/non-state-threading actor message) that captures a `HandleScoped(#process)` value warns |
| **BT-2757** | Announcement payload sendability (reuses the message-arg machinery); undeclared-handle companion lint nudging `native:` Object classes with instance behaviour and no scope |
| **BT-2758** | End-to-end diagnostic + hover integration tests; `docs/beamtalk-language-features.md` "Sendability Tiers" section; ADR 0103 flipped Proposed→Accepted with an Implementation Tracking table |

## Notable design decisions

- **`ClassInfo` field, not a side-map** — `add_from_beam_meta` round-trips through `Vec<ClassInfo>`, so `handle_scope` belongs there. Propagated to the generator (`build_stdlib`) + regenerated `generated_builtins.rs` (verified identical to the hand-edit) and all literal sites.
- **Block-capture typing** — `CapturedVar.defined_at` is a definition span, not a `type_map` key; the captured set stays authoritative but each type is recovered from the variable's *use* inside the block (which the type checker records).
- **Single tier core** — every boundary check (message args, `spawnWith:`, announcements, block captures, hover) calls the one `tier_of`, so tiers cannot drift.
- **`#node` silent in v1** — no static remoteness knowledge (ADR 0079 is local-only); remote-`#node` grading is recorded as deferred/blocked on a future cluster-registration ADR.
- **`__beamtalk_meta`** emits `handle_scope` only when declared, keeping meta output byte-identical for the vast majority of classes (no codegen snapshot churn).

## Verification

- `cargo build` / `clippy` / `fmt` clean across the workspace
- beamtalk-core: 4448 unit + integration tests pass (incl. new sendability, parser, hierarchy, validator, and hover tests)
- stdlib language-feature tests: 223 pass · BUnit: 0 failed · `test-docs` passes
- Pre-push CI hook (Erlang build, spec validation, Dialyzer) passed before push

No runtime or codegen output-semantics change — metadata + diagnostics + one keyword only.

https://claude.ai/code/session_01FP3kLDrHLrBUMY3xEd3gFP

---
_Generated by [Claude Code](https://claude.ai/code/session_01FP3kLDrHLrBUMY3xEd3gFP)_


jamesc:
Thanks for the review. Also merged `main` (the ADR 0104 siblings BT-2749/BT-2750 landed there) and propagated `handle_scope` to the new `ClassInfo`/`ClassDefinition` test fixtures so CI compiles again.

On the three points:

1. **Cross-package `handle_scope` dropped in the BEAM port** — fixed. `parse_class_info_from_meta_term` now reads the key back from the meta map (`map_get(m, "handle_scope").and_then(term_to_atom)`), since a handle scope is a declaration and can't be structurally re-derived. A binary-only dependency that declares `handleScope:` keeps its tier across package boundaries now.

2. **Generic field annotations lose `type_args` in Value composition** — confirmed real (advisory-only, fails silent). The root cause is at the storage layer: `ClassInfo.state_types` keeps only the name string (`"List(Port)"`), so the generic args are gone before `tier_of` runs — more than an inline fix. Tracked as **BT-2770**.

3. **Misplaced `handleScope:` → confusing DNU** — agreed, a targeted parse error would be friendlier. Tracked as **BT-2771**.

---

jamesc:
All five points from this pass addressed in `ff6b55d`:

- **`handleScope:` on Actor → wrong tier** — fixed. The declared scope is now consulted only inside the `Object` arm of the kind match, so an `Actor subclass: … handleScope: #process` stays `SendableRef` (matching the `check_handle_scope_on_object` advisory that already flags the stray declaration). Test added.
- **`Tier::meet` under-warns intersections** — fixed. Intersections now use `join` (most-hazardous member), so `Pid & Port` derives `HandleScoped(#process)` instead of masking to `SendableRef`. Removed the now-unused `meet`. Test added.
- **Parser doesn't consume the bad token** — fixed. A non-symbol after `handleScope:` is now consumed (guarded on same-line) so class-body parsing recovers cleanly, mirroring the `native:` error path.
- **Nit: `.app.src` omits `handle_scope`** — documented the intentional omission (the authoritative channel is `__beamtalk_meta`, read back by the compiler-port; no stdlib class declares a scope today).
- **Nit: missing test assertion** — added a `handle_scope: None` assertion to `test_generate_class_entry_emits_correct_structure`.

`just test-rust`, clippy, and fmt are green locally.

---

jamesc:
Addressed the sendability-side points in `5f740bf`:

- **#2 Misplaced `handleScope:`** — now emits a targeted parse error ("must appear in the class header, before state or method declarations") and consumes the clause for recovery, instead of falling through to a stray-keyword message. Test added; closes BT-2771.
- **#3 Class span excludes the clause** — the `handleScope:` span is now merged into the class-definition end span, so a scope-only class body spans its declaration for hover/diagnostic ranges.
- **Nit: companion lint ignores inherited scope** — switched to the inherited `ClassHierarchy::handle_scope` accessor, so a `native:` subclass that inherits a scope from a parent is no longer nudged.

Left as-is:
- **#1 `withTimeout:` Actor-override test** and the inference.rs comment nit — that's BT-2751's code (merged from `main`), out of scope for this PR's diff.
- **Nit: parenthesized `(self)`** in `send_crosses_process_boundary` — negligibly rare and advisory-only; not worth the special-case.

`just test-rust`, clippy, fmt green locally.

---

PR #2884, opened 2026-07-09
test(codegen): add Rust unit tests for BT-2703 eachWithIndex:/do:separatedBy: desugar

jamesc:
## Summary

Adds 8 Rust unit tests covering the previously-untested paths in `enumeration_ops.rs` (the BT-2703 `eachWithIndex:` / `do:separatedBy:` desugar). Coverage on that file was 85.2% (144/169 lines); all 25 uncovered lines are now exercised.

### What's tested

**`eachWithIndex:` desugar paths:**
- Actor field mutation → desugars to `lists:foldl` with `{'nil', element(2, Fold)}` re-projection
- Actor local mutation → same desugar, local threaded through fold accumulator
- Wrong-arity block (1-arg for a 2-arg op) → falls through to `Collection.bt` dispatch
- Pure block (no mutations) → falls through to `Collection.bt` dispatch
- Value-type method mutation → desugars to `lists:foldl` but without the actor re-projection

**`do:separatedBy:` desugar paths:**
- Actor element-block mutation → desugars with "is-first" flag accumulator + re-projection
- Actor separator-block-only mutation → same desugar triggered by the separator block
- Both blocks pure → falls through to `Collection.bt` dispatch

### Why Rust unit tests rather than BUnit tests

The BUnit runtime tests for these selectors already exist in `stdlib/test/collection_mutations_test.bt`, but they run through `beamtalk test` and are not instrumented by `cargo llvm-cov`. The codegen paths in `enumeration_ops.rs` can only be reached via `cargo test`, hence the Rust unit tests using the `codegen()` helper.

### Notes on assertions
The generated module metadata always records `'eachWithIndex:'` / `'do:separatedBy:'` in the `methodXref.sends` map regardless of whether the desugar fired, so negative assertions target the `lists:foldl` presence/absence rather than the selector string.

---
_Generated by [Claude Code](https://claude.ai/code/session_013CPUNxAqzuCVYQp2Py213q)_

### DIFF
--- a/crates/beamtalk-core/src/codegen/core_erlang/control_flow/list_ops/tests.rs
+++ b/crates/beamtalk-core/src/codegen/core_erlang/control_flow/list_ops/tests.rs
@@ -1575,6 +1575,14 @@ fn test_sort_pure_generates_beamtalk_list_sort_with() {
 // correctly threaded.  Non-mutating blocks fall through to the `Collection.bt`
 // dispatch.  The selector still appears in the generated method metadata, so
 // negative assertions must target the *dispatch body*, not the whole module.
+//
+// Fall-through assertions below match on the literal generated variable name
+// `_items1`: every test in this section uses a single-statement `run: items`
+// method, so the collection parameter is always the first extracted arg and
+// always gets the deterministic name `_items1` (param name + arg-position
+// counter). This is stable for these snippets but is an artifact of codegen's
+// naming scheme, not a documented contract — if that scheme changes these
+// assertions need to change with it.
 
 #[test]
 fn test_each_with_index_field_mutation_desugars_in_actor() {
@@ -1661,6 +1669,10 @@ fn test_each_with_index_value_type_desugars_without_reprojection() {
     assert!(
         code.contains("'lists':'foldl'"),
         "eachWithIndex: Value type: should still desugar to lists:foldl. Got:\n{code}"
+    );
+    assert!(
+        code.contains("'maps':'put'('__local__acc'"),
+        "eachWithIndex: Value type: should thread 'acc' through the fold accumulator. Got:\n{code}"
     );
     assert!(
         !code.contains("{'nil', call 'erlang':'element'(2,"),
@@ -1773,6 +1785,9 @@ fn test_do_separated_by_non_literal_callable_falls_through() {
     // this scenario's field mutation would hit if actually compiled.
     let src = "Actor subclass: Ctr\n  state: total = 0\n\n  run: items =>\n    blk := [:x | self.total := self.total + x]\n    items do: blk separatedBy: [nil]\n";
     let code = codegen(src);
+    // Bracket deliberately left open (no trailing `]`) here, unlike the
+    // eachWithIndex: assertion above: do:separatedBy: passes two arguments
+    // (`[Blk, fun () -> 'nil']`), so `[Blk]` alone would never match.
     assert!(
         code.contains("'send'(_items1, 'do:separatedBy:', [Blk"),
         "do:separatedBy: non-literal callable: should dispatch via message send. Got:\n{code}"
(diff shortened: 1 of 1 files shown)


jamesc:
Addressed all suggestions/nits from the review in 704752d (re-signed as 12c10f6 after rebase):

- Strengthened `test_each_with_index_local_mutation_desugars_in_actor` to assert `'maps':'put'('__local__acc'` (local threading, not just fold presence).
- Strengthened `test_do_separated_by_separator_block_mutation_desugars_in_actor` to assert `'maps':'put'('count'` (field threading in the nested separator case).
- Added `test_each_with_index_degenerate_param_names_falls_through` for the `[:x :x | …]` guard.
- Added `test_each_with_index_non_literal_callable_falls_through` and `test_do_separated_by_non_literal_callable_falls_through` for the non-literal-block fallback.
- Added `test_do_separated_by_value_type_desugars_without_reprojection` mirroring the existing `eachWithIndex:` Value-type coverage.

All 12 tests in this section pass, plus the full `cargo test` suite (4564 passed) and `just clippy`/`just fmt` are clean.

---

jamesc:
Addressed this second round in 058b53e:

- **Weak positive assertions (suggestion 1)**: tightened all 4 fall-through tests' positive assertions to `code.contains("'send'(_items1, '<selector>:', [")`, which only matches the actual dispatch call site — not the method metadata, which unconditionally records the selector regardless of whether the desugar fired.
- **Non-literal callable state-mutation gap (suggestion 2)**: dug into this and confirmed it's worse than "silently lost" — the generated Core Erlang fails `erlc` outright with `unbound variable 'State1' in dispatch/4`. Verified this is a **pre-existing, general bug unrelated to BT-2703/enumeration_ops.rs** — it reproduces with a plain `blk value: item` direct call, no `eachWithIndex:`/`do:separatedBy:` involved (root cause: the shared state-version counter used by `self.field :=` codegen leaks out of a `Block` literal's own function scope). Filed [BT-2792](https://linear.app/beamtalk/issue/BT-2792) with a full repro and root-cause analysis, and referenced it in the two non-literal-callable test comments rather than expanding this test-only PR's scope to fix production codegen.
- **Missing non-literal separator test (suggestion 3)**: added `test_do_separated_by_non_literal_separator_falls_through`.
- **Vestigial state field (nit)**: renamed `state: total = 0` → `state: dummy = 0` in the two Value-type tests to match the existing convention in this file (only `acc`, a local, is actually exercised).

All 13 tests in this section pass, plus the full suite (4565 passed) and `just clippy`/`just fmt`.

---

jamesc:
Addressed this round in 5173059:

- Both Value-type tests (`test_each_with_index_value_type_desugars_without_reprojection`, `test_do_separated_by_value_type_desugars_without_reprojection`) now assert `'maps':'put'('__local__acc'`, so a desugar that drops the accumulator update would actually fail these tests instead of silently passing.
- Documented the `_items1` naming convention in the section header comment — it's deterministic for these single-statement snippets but is a codegen artifact, not a guaranteed contract.
- Added a one-line comment on the `do:separatedBy:` non-literal-callable assertion explaining why its bracket is deliberately left open (`[Blk` vs `[Blk]`) — that selector takes two arguments.

All 13 tests pass, plus the full suite (4565 passed), clippy, and fmt.

---

jamesc:
Thanks for the PASS — closed out the two remaining low-risk suggestions in a0436fb:

- `test_each_with_index_zero_arity_block_falls_through`: covers the 0-arg case distinct from the existing 1-arg wrong-arity test (the guard is `!= 2`, not specifically 1-arg).
- `test_do_separated_by_separator_with_param_falls_through`: covers the `!separator_block.parameters.is_empty()` guard directly.

Now 15 tests in this section, all passing, plus the full suite (4567 passed), clippy, and fmt.

Separately, the earlier "Claude BeamTalk Review — did not complete" failure on this PR was a GitHub Actions workflow-validation quirk (unrelated to this branch's diff — the workflow file is byte-identical to `main`); I re-ran the failed job and it came back PASS.

---

jamesc:
Good catch — verified against the codegen source (`gen_server/dispatch.rs`: parameters are registered via `fresh_var` before any body statement is generated, per the "Register parameters BEFORE generating body" step) and fixed the rationale in 42d0a5a. The "single-statement" claim was indeed wrong; the real reason is that parameter binding is always the first `fresh_var` call for the method, independent of body statement count.

Full suite (4567 passed), clippy, and fmt all clean.

---

PR #2899, opened 2026-07-10
Codegen: thread self.field mutations through stored/opaque blocks (BT-2797)

jamesc:
## Summary

Follow-up to BT-2792 (which added a compile-time safety-net diagnostic for field mutations in blocks the compiler couldn't safely thread state through). This PR makes the two most common such shapes actually *work* instead of hitting that diagnostic:

- **Same-method local var**: `blk := [:x | self.total := self.total + x]. ... blk value: item` later in the same method. A new per-method pre-scan (`prescan_tier2_local_vars`) proves every later reference to the local is a safe `value`/`value:`/etc. call before promoting it to the Tier 2 stateful protocol — if the block could escape (returned, reassigned, passed elsewhere), it's left unpromoted and still hits BT-2792's diagnostic.
- **Cross-method field-stored blocks** (the main real-world motivator — a block assigned to `self.field` in one method, invoked via `self.field value:` from a different one, e.g. a deferred callback). Static per-method tracking can't see across methods, so every `self.field value(:...)` call site now runtime-discriminates Tier 1 vs Tier 2 via `erlang:is_function/2`, generalizing the BT-909 precedent (previously only used for Erlang FFI interop) to Beamtalk-level block calls.

During development this also caught and fixed a vacuous-truth bug in the same-method safety check: "no unsafe use found" is trivially true when a variable is simply never referenced again (e.g. as a method's last statement, letting the block value implicitly escape as the return value) — the check now requires proof of at least one actual *safe* use, not just the absence of an unsafe one.

`valueWithArguments:` turned out to need a materially different fix (it's a stdlib `@primitive` compiled once into an actor-agnostic module, unlike `value`/`value:` which are call-site-spliced `@intrinsic`s) — filed as BT-2803 rather than folded in here.

## Test plan

- [x] `cargo test -p beamtalk-core --lib` — 4601 passed
- [x] `just clippy` / `just fmt` — clean
- [x] `just test-bunit` — 2685 tests, 0 failed (new `tier2_field_stored_block_test.bt` covers the cross-method field-stored case end-to-end at runtime)
- [x] `just ci` — full suite clean (build, lint, unit/BUnit/runtime/integration/MCP/parity/REPL-protocol tests, corpus check, surface-drift check)
- [x] New codegen regression tests cover: the original BT-2792 repro now compiling with correct Core Erlang shape, the cross-method field-stored case's generated Core Erlang shape, the vacuous-truth escape-hatch fix, a non-literal block argument to a self-send still correctly erroring at compile time, and no regression to the zero-cost pure-block fast path

Linear: BT-2797

---
_Generated by [Claude Code](https://claude.ai/code/session_01W7etfDiPFnHipA4WbSqeHA)_


jamesc:
Auto-merged by daily routine after passing CI + CodeRabbit + Copilot gates.

---

PR #2907, opened 2026-07-11
Codegen: fix three Tier 2 stored/opaque block bugs (BT-2807, BT-2808, BT-2809)

jamesc:
## Summary

Fixes three codegen bugs discovered during BT-2797's review (PR #2899), all in the "field-stored/opaque Tier 2 block" area — a block stored in an instance field or local variable, mutating actor state, invoked later via `value:`/`value:value:`/cascade.

- **BT-2807** — a bare `self.field value:` with no other field mutation in the same nested block (e.g. inside `ifTrue:`) didn't trigger state threading, so a genuinely Tier 2 stored block crashed with `badarity` or silently lost its mutation. `block_facts::analyze_block` now tracks a `has_field_value_call` fact and folds it into `has_state_effects()`.
- **BT-2808** — `blk value: x; value: y` cascades on a Tier 2 local var/param/field-stored block were rejected at compile time. The parser folds a cascade's first message into `Cascade.receiver` as a whole `MessageSend`, which the original safety scan didn't account for. Added `normalize_cascade` to decompose the true receiver + full message list, and `generate_tier2_cascade_doc` to thread state sequentially through each cascaded send.
- **BT-2809** — a field-stored block whose only mutation is a captured local (not a field write) compiled and ran without crashing, but silently corrupted actor state when invoked from a different method than the one that assigned it (stale definition-time value returned forever, stray `'__local__'` key leaking into persistent gen_server state). Now a clear compile-time diagnostic instead.

A 4th commit addresses two issues found in this PR's own code review:
- `block_facts.rs`'s cascade handling for BT-2807 only checked the cascade's *first* message for a `self.field value:` send, missing the identical pattern as a later cascaded message.
- BT-2809's diagnostic double-fired (two separate errors) for a block with both a field write and a captured-local mutation; now emits just the one.

Also filed **BT-2810** as a follow-up: a systematic test matrix for the field-stored/opaque Tier 2 block axis, since coverage there was reactive (one hand-written regression test per bug found in review) rather than systematic like the existing `mutation_matrix_test.bt` (BT-1480).

## Test plan

- [x] `just test` (Rust unit tests, stdlib bootstrap tests, BUnit, Erlang runtime tests) — all green, no regressions
- [x] `just clippy` / `just fmt` — clean
- [x] New Rust unit tests for each bug (`block_facts.rs`, `block_analyzer.rs`, `control_flow.rs` codegen shape tests)
- [x] New end-to-end BUnit tests exercising each fix at runtime (`tier2_subexpr_and_conditional_block_test.bt`, `tier2_field_stored_block_test.bt`)
- [x] Manually verified BT-2809's diagnostic surfaces correctly via `beamtalk build` with source span + hint
- [x] 8-angle code review (correctness, removed-behavior, cross-file, reuse, simplification, efficiency, altitude, conventions) — 2 correctness gaps found and fixed; remaining findings are pre-existing/architectural cleanup opportunities filed as review notes, not blocking

---
_Generated by [Claude Code](https://claude.ai/code/session_01BHhF87zXFhcAjJbetpCdQG)_

### DIFF
--- a/crates/beamtalk-core/src/codegen/core_erlang/gen_server/methods.rs
+++ b/crates/beamtalk-core/src/codegen/core_erlang/gen_server/methods.rs
@@ -13,9 +13,9 @@ use super::super::document::{Document, INDENT, join, leaf, line, nest};
 use super::super::selector_mangler::safe_class_method_fn_name;
 use super::super::{CodeGenContext, CodeGenError, CoreErlangGenerator, Result, block_analysis};
 use crate::ast::{
-    Block, ClassDefinition, ClassKind, Expression, Identifier, Literal, MapPair, MessageSelector,
-    MethodDefinition, MethodKind, Module, ParameterDefinition, StateDeclaration, TypeAnnotation,
-    TypeParamDecl, WellKnownSelector,
+    Block, CascadeMessage, ClassDefinition, ClassKind, Expression, Identifier, Literal, MapPair,
+    MessageSelector, MethodDefinition, MethodKind, Module, ParameterDefinition, StateDeclaration,
+    TypeAnnotation, TypeParamDecl, WellKnownSelector,
 };
 use crate::docvec;
 use crate::unparse::unparse_method_display_signature;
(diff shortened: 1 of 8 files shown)


jamesc:
Thanks for the review. Went through all three:

**1. Multiple `CapturedVariable` diagnostics per block** — this matches the existing `Field` arm's behavior exactly (which already fires once per field-write mutation, not deduped by field name). The `bt2809_mixed_field_and_captured_local` test's concern was specifically about the *cross-kind* case (a Field diagnostic + a CapturedVariable diagnostic for the same block, describing the same underlying defect) — not about deduping multiple distinct captured variables, which is legitimate: each names a genuinely separate offending variable, same as multiple field writes would. Deduping only `CapturedVariable` and not `Field` would be an inconsistency, so leaving this as-is.

**2. Intermediate `_CascResultN` bindings** — verified this doesn't actually produce an erlc warning: `fresh_temp_var` always underscore-prefixes generated names (`_CascResult1`, `_CascResult2`, ...), which is this repo's own documented convention for suppressing unused-variable warnings (`docs/development/erlang-guidelines.md`: "Prefix unused variables with underscore"). Confirmed with a manual 3-message cascade build — no warnings. Binding-then-discarding an intermediate tuple-unpack result to sequence the next `apply` is also the same idiom already used elsewhere in this file (e.g. `_T2Discard` in the non-last bare `Tier2ValueCall` case), so this is consistent with existing style, not dead code.

**3. `normalize_cascade` duplicated between `methods.rs` and `block_facts.rs`** — agreed, flagged this exact duplication myself in review (plus a third copy already existing in `expressions.rs`'s `generate_cascade`, which `methods.rs`'s copy already mirrors). Deferring — unifying it means either normalizing `Cascade`'s AST shape at parse time or exposing one shared helper both `semantic_analysis` and `codegen` can call, which is a bigger, separate change than these three bug fixes warrant. Left as a known cleanup opportunity rather than scope-creeping this PR.

---

PR #2909, opened 2026-07-11
Follow-up issues BT-2803, BT-2804, BT-2805

jamesc:
## Summary

Three independent follow-up issues from the BT-2791 typing-cleanup / ADR 0105 epics:

- **[BT-2805](https://linear.app/beamtalk/issue/BT-2805)** — `beamtalk_recheck.erl`'s `retyped_fallback/1` picked the *first* retyped field by list order when 2+ fields were retyped in the same reload, silently blaming a possibly-wrong field in the recheck finding. Now surfaces the other equally-plausible retyped fields and renders the ambiguity explicitly in the finding's `note` instead of asserting a single field with false confidence (per human-selected option (b) on the issue).
- **[BT-2804](https://linear.app/beamtalk/issue/BT-2804)** — `analyse_full`/`analyse_full_with_natives` threaded 9–12 positional parameters (several bare, transposition-prone bools) through ~6 public wrappers. Replaced with a single `AnalysisContext` struct whose `Default` reproduces today's conservative `ModuleOnly` behavior, and merged the two functions into one. No behavior change — public wrapper signatures are untouched.
- **[BT-2803](https://linear.app/beamtalk/issue/BT-2803)** — `Block>>valueWithArguments:` was a bare `@primitive` compiled once into a shared, actor-agnostic stdlib module, so it always did `erlang:apply(Self, Args)` — badarity-crashing when `Self` was a Tier 2 (stateful) block needing a trailing state argument. Converted to a call-site-intercepted `@intrinsic` (mirroring `value`/`value:`'s BT-2797 precedent), with a new runtime Tier 1/Tier 2 discrimination generalized to a runtime-length argument list (`is_function(Fun, length(Args))` vs `length(Args)+1`), threading the calling actor method's live state for the Tier 2 branch.

## Key changes

- `runtime/apps/beamtalk_workspace/src/beamtalk_recheck.erl` / `beamtalk_recheck_tests.erl`
- `crates/beamtalk-core/src/semantic_analysis/mod.rs` — new `AnalysisContext` struct
- `crates/beamtalk-core/src/codegen/core_erlang/{intrinsics.rs,gen_server/methods.rs,primitives/block.rs}` — `valueWithArguments:` intrinsic conversion, plus shared `hoist_open_scope_receiver`/`hoist_open_scope_argument` helpers extracted during review to remove duplication
- `stdlib/src/Block.bt`, `stdlib/test/tier2_field_stored_block_test.bt` + fixture — new BUnit coverage for a Tier 2 field-stored block invoked via `valueWithArguments:` across methods

## Testing

- `cargo test -p beamtalk-core --lib` (4629 passed), `cargo clippy` clean
- `rebar3 eunit --module=beamtalk_recheck_tests` (52 passed)
- `just test-stdlib` (223 passed), `just test-bunit` (full suite green)
- Full pre-push CI (clippy, rustfmt, erlfmt, dialyzer, spec validation) green

## Follow-up filed

Adversarial code review surfaced a pre-existing, systemic gap (not introduced here, but newly inherited by `valueWithArguments:`'s conversion): structural intrinsics (`value`, `value:`, `valueWithArguments:`, etc.) are unreachable via `perform:withArguments:` — confirmed the same failure already exists for `value`/`value:value:` at `main`. Filed [BT-2812](https://linear.app/beamtalk/issue/BT-2812) to track a proper fix, and left a code comment at the mechanism (`generate_primitive`'s fallback in `codegen/core_erlang/mod.rs`).

---
_Generated by [Claude Code](https://claude.ai/code/session_01TsfWKEKiP6js3RTwgbBE5a)_


jamesc:
Fixed the `is_self_field_value_send` gap in 8ea0a259 — `valueWithArguments:` now recognized alongside `value`/`value:`/etc, matching your suggested one-liner exactly.

While verifying it end-to-end (nested `do:` block, field-stored Tier 2 block invoked via `valueWithArguments:`), I confirmed the fix closes the *predicate* gap, but the deeper "state has nowhere to go" crash you described also reproduces identically for plain `value:` in the same nested position (pre-existing, not introduced by BT-2803 or BT-2807 — `do:`/`timesRepeat:` block bodies compile through a separate path with no Tier 2 tuple-unpacking). So this brings `valueWithArguments:` to exact parity with `value:`'s existing behavior rather than fully closing the nested-block gap for either. Not filing a separate BT issue since it's the same class of limitation already implicitly accepted for the sub-expression-position case (see `try_generate_block_value_keyword`'s BT-2797 comment) — happy to file one if you'd rather track it explicitly.

Also fixed a real CI break this surfaced: `beamtalk_primitive_tests.erl`'s `send_block_valueWithArguments_test_` (reflective dispatch via `beamtalk_primitive:send/3`) broke because the `@primitive` → `@intrinsic` conversion left `blockValueWithArguments`'s compiled body as an unreachable placeholder — gave it a real fallback body (`erlang:apply(Self, Args)`, same as the old primitive) since, unlike `value`/`value:`, it never needs state threading in that fallback (a Tier 2 block can't reach it — see BT-2812 for the broader, pre-existing `value`/`value:` gap this doesn't touch).

---

jamesc:
Addressed in 40051594:
- **S1**: `generate_tier2_value_call_doc` now fails loudly (`CodeGenError::Internal`) instead of silently falling through if `valueWithArguments:` ever had no argument expression.
- **S2**: `field_change_note/3`'s `added`/`removed` clauses now match any `AmbiguousWith` instead of only `[]`, removing the `function_clause` crash risk.
- **N1**: strengthened the local-var fast-path test to positively assert the `is_function/1` guard is present on the hoisted receiver.

N2 (comment length) left as-is — agreed it's pure style with no behavior impact.

---

PR #2928, opened 2026-07-12
Extend isKindOf: control-flow narrowing to post-guard early-return shapes (BT-2825)

jamesc:
## Summary

Linear: https://linear.app/beamtalk/issue/BT-2825

`(x isKindOf: Foo) ifFalse: [^default]` — the guard-and-early-return idiom already handled for `isNil` (BT-2049) — previously left `x` unnarrowed for the rest of the method, because `apply_early_return_narrowing` only recognised `isNil` checks and only the `ifTrue:`/`ifTrue:ifFalse:` shapes (never a solo `ifFalse:`). This closes that gap for `isKindOf:` and `class =:=` guards.

- **`(x isKindOf: Foo) ifFalse: [^default]`** now narrows `x` to `Foo` for the rest of the method (previously required an `@expect type` escape hatch or produced a false type-mismatch warning downstream).
- **`(x isKindOf: Foo) ifTrue: [<diverge>]`** now narrows `x` to the complement (`T \ Foo`) for the rest of the method — the mirror image.
- **Protocol collapse**: when the pre-guard type is a Protocol (e.g. `Printable`) and the `isKindOf:` target is an unrelated concrete class, the narrowed type collapses the resulting `Intersection` down to the bare concrete class, so it satisfies a concrete-class-typed assignment (`items :: List := coll`) without an `@expect type` escape hatch.
- **`class =:=`'s false branch stays unnarrowed**, exactly as before (BT-2744) — a `Character` subclass could still be "not exactly Integer" yet satisfy `isKindOf: Integer`, so narrowing it away would be unsound.

### Key changes

- `crates/beamtalk-core/src/semantic_analysis/type_checker/inference.rs`: factored `refine_class_narrowing`'s narrowing math into a diagnostic-free `compute_class_narrowing` helper shared with `apply_early_return_narrowing` (which was previously `isNil`-only and had no solo-`ifFalse:` shape at all). `refine_class_narrowing` keeps emitting the "comparison can never be true" hint (unchanged); the new post-guard path does not re-emit it, since the guard expression was already fully type-checked.
- `crates/beamtalk-core/src/semantic_analysis/type_checker/tests/narrowing_post_guard.rs`: 6 new tests covering the union-narrows-after-`ifFalse:` case, the Protocol-collapse case (the exact `prompt_renderer.bt` shape from the issue), the `ifTrue:`-diverge complement case, a regression lock-in for `isKindOf:` inside an `ifTrue:` branch, and two soundness negatives (a non-diverging guard must not narrow; `class =:=`'s false branch must stay unnarrowed even post-guard).
- `docs/beamtalk-language-features.md`: documented the new post-guard `isKindOf:` patterns in the Control Flow Narrowing table.

### Tests

- `cargo test -p beamtalk-core --lib`: 4666 passed, 0 failed (includes the 6 new BT-2825 tests plus the full existing `narrowing_post_guard` / `type_checker` suite — 985 type_checker tests, all green).
- `cargo clippy -p beamtalk-core --all-targets -- -D warnings`: clean.
- `cargo fmt --check`: clean.
- `git push` ran the repo's lint gate (clippy + fmt) clean on push.

---

### DIFF
--- a/crates/beamtalk-core/src/semantic_analysis/type_checker/inference.rs
+++ b/crates/beamtalk-core/src/semantic_analysis/type_checker/inference.rs
@@ -4009,6 +4009,13 @@ impl TypeChecker {
             // argument is the false block. If it diverges, execution
             // reaching the next statement proves the guard's test held,
             // so the variable narrows to the *true*-branch type.
+            //
+            // For `isKindOf:` this is `compute_class_narrowing`'s resolved
+            // `true_type` (set above). For plain `isNil` checks, `info` was
+            // never routed through `compute_class_narrowing` — `true_type`
+            // instead comes straight from `detect_narrowing`'s `isNil` rule,
+            // which sets it to `InferredType::known("UndefinedObject")`
+            // (see `narrowing/rules/is_nil.rs`).
             let Some(Expression::Block(false_block)) = arguments.first() else {
                 return;
             };
--- a/crates/beamtalk-core/src/semantic_analysis/type_checker/inference.rs
+++ b/crates/beamtalk-core/src/semantic_analysis/type_checker/inference.rs
@@ -2464,18 +2464,70 @@ impl TypeChecker {
     ///
     /// `detect` only sees the AST, so it records the tested class name in
     /// `class_test` and leaves `true_type` provisional. Here we resolve the
-    /// variable's current type and route the **true** branch through the
-    /// hierarchy-aware `intersect(current, C)` for *both* idioms, and the
-    /// **false** branch through `difference(current, C)` for `isKindOf:`
-    /// *only* — see `ClassTestKind` for why `class =:=`'s false branch must
-    /// stay unnarrowed.
+    /// variable's current type and delegate to the diagnostic-free
+    /// [`Self::compute_class_narrowing`] (BT-2825 factored this out so
+    /// [`Self::apply_early_return_narrowing`] can reuse the same math without
+    /// re-emitting the "comparison can never be true" hint below for a guard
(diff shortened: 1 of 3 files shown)


jamesc:
Addressed the suggestion in `e189994`: added `bt2825_is_nil_if_false_diverge_narrows_to_nil` locking in the `x isNil ifFalse: [^...]` post-guard behavior, plus a doc row in the Control Flow Narrowing table. The other two nits (a `class =:= ... ifTrue:ifFalse:` shape test, and calling out the protocol-collapse fix explicitly) are low-risk/documentation-only and left as-is — the underlying code path is already covered by the existing `ifTrue: [<diverge>]` test.

---

jamesc:
These three are style/documentation nits under a PASS verdict with no functional impact — leaving as-is rather than triggering another CI round:
- `&mut self` on `apply_early_return_narrowing`: correct that the body doesn't mutate `self`, but it's called from `infer_stmts` which already holds `&mut self`, so this is harmless over-widening, not a defect.
- The `isOk`/`isError`/`singleton_eq` scope exclusion is intentional (this issue's scope is `isNil`/`isKindOf:` only) — a real follow-up if someone wants to extend post-guard narrowing further, not a gap in this PR.
- The `[^...]` docs notation ambiguity for the new rows: the prose paragraph immediately below the table already covers this for `ifTrue:`, and the new `ifFalse:`/`ifTrue:ifFalse:` rows share the same diverge-detection code path (`block_diverges`), so the same caveat applies without needing per-row restatement.

---

PR #2929, opened 2026-07-12
Some stdlib methods declare callback params as bare Block instead of parameterized Block(...) (BT-2826)

jamesc:
Linear: https://linear.app/beamtalk/issue/BT-2826/some-stdlib-methods-declare-callback-params-as-bare-block-instead-of

## Summary

`String>>collect:` and `Actor>>onExit:` declared their callback params as bare `Block` instead of `Block(...)`, forcing `typed` callers to add an `@expect type` workaround plus a rebind to recover a concrete block-param type. Fixed both, plus completed the acceptance criteria's audit of the rest of the stdlib for the same pattern.

## Changes

- Parameterized the two methods named by the issue: `String>>collect:` → `Block(String, String)`, `Actor>>onExit:` → `Block(Object, Object)`.
- Audit turned up and fixed the same bare-`Block` pattern in: `String>>do:/select:/reject:`, `Block>>on:do:`, `List>>groupBy:`, `Stream>>from:by:/select:/collect:/reject:/do:/inject:into:/detect:/anySatisfy:/allSatisfy:`, `Integer>>to:do:/to:by:do:`, `Interval>>do:`.
- Deliberately left bare (with an explanatory `//` comment) where parameterizing breaks something real:
  - `Binary>>do:` — String overrides it with grapheme-`String` elements, not bytes; parameterizing Binary's would make the override incompatible under the invariant override check.
  - `Announcer>>when:do:`/`when:doOnce:` — the announcement's real type depends on the runtime `aClass` argument, not a fixed static bound. Parameterizing to `Object` broke real code accessing event-specific fields (caught by `test/system_announcements_test.bt`).
- Gave `Stream>>inject:into:` a local generic accumulator type param (`A`), matching `Collection`'s existing convention, instead of a fixed `Object` bound — the fixed bound broke `count + 1`-style accumulation (caught by `test/file_stream_test.bt`).
- Added a Rust regression test (`crates/beamtalk-core/src/semantic_analysis/type_checker/tests/bt2826_block_param_types.rs`) verifying both named methods now resolve block params to concrete (non-Dynamic) types with no DNU and no "unannotated parameter" warning.

Filed [BT-2834](https://linear.app/beamtalk/issue/BT-2834/audit-zero-argambiguous-arity-bare-block-callback-params-in-stdlib) to track the remaining zero-arg/ambiguous-arity bare `Block` methods (`ifTrue:`/`ifNil:`/`whileTrue:`/etc.) as a separate, higher-risk follow-up touching more foundational classes.

## Test plan

- [x] `cargo test -p beamtalk-core --lib` — 4662 passed
- [x] `cargo clippy --workspace -- -D warnings` — clean
- [x] `cargo fmt --check` — clean
- [x] `beamtalk build-stdlib --warnings-as-errors` — clean (0 override-compatibility warnings)
- [x] `just test-bunit` — 2720 tests, 2718 passed, 0 failed
- [x] `just test-stdlib` — 223 tests, 223 passed
- [x] New regression test for both named methods (`bt2826_string_collect_infers_string_block_param`, `bt2826_actor_on_exit_infers_object_block_param`)

---


jamesc:
Addressed the zero-arg-handler suggestion in `c16d268`: added `bt2826_on_do_zero_arg_handler_in_typed_context_no_diagnostics`, confirming `on: Error do: [0]` type-checks cleanly in a `typed` class (block-param arity checking is flexible, so this was never actually broken — just untested). The two nits (sparser test coverage for the other ~13 signatures, and `Stream>>collect:`'s pre-existing bare `Stream` return type) are left as follow-up scope, not required for this fix.

---

jamesc:
Agreed this is a real gap but out of scope for this PR (as noted, not a regression — this PR improves it directionally). Tracked as [BT-2835](https://linear.app/beamtalk/issue/BT-2835/streamcollect-return-type-doesnt-propagate-the-transforms-return-type).

---

PR #2931, opened 2026-07-12
Thread selector/class breadcrumb into self-dispatch reraise (BT-2822)

jamesc:
## Summary

Follow-up from BT-2816 ("Self-send forwarding into a method that raises `self error:` crashes with `function_clause`"). BT-2816 fixed the self-dispatch error branch to destructure `safe_dispatch/3`'s caught-exception triple and route it through `beamtalk_exception_handler:'reraise'/3`, but that's the context-free variant. The cross-actor call boundary (`beamtalk_actor:sync_send_remote/3`) already does one step better: it calls `reraise/4` with a `Context` map (`#{selector => Selector, class => Class}`), which feeds `wrap_raw/2`'s location-prefixing (`ClassName>>selector: ...`).

This PR threads that same breadcrumb into `generate_self_dispatch_error_clause` and its call sites, so a raw Erlang error escaping a forwarded self-send gets the same location prefix as the cross-actor equivalent, instead of losing it via the context-free `reraise/3`.

Known gap: the breadcrumb uses the compile-time defining class, not the runtime receiver class under inheritance — tracked separately as follow-up **BT-2833**.

## Test plan

- [x] `just build` — clean build
- [x] `just clippy` — passes
- [x] `just fmt-check` — passes
- [x] Pre-push hook (clippy, dialyzer, full build+stdlib) — all green
- [x] New Rust codegen test in `crates/beamtalk-core/src/codegen/core_erlang/tests/dispatch.rs`
- [x] New BUnit test `stdlib/test/actor_self_send_raw_error_test.bt` + fixture, asserting the located message text
- [x] Existing BT-2816 regression tests (`actor_self_call_error_test.bt`, `test_self_call_error_branch_handles_both_error_shapes`) unaffected

## References

* Linear: [BT-2822](https://linear.app/beamtalk/issue/BT-2822/codegen-thread-selectorclass-breadcrumb-into-self-dispatch-reraise-bt)
* Follow-up filed for the inheritance gap: BT-2833
* Related: BT-2816, BT-2704, BT-2705

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01HLRMHdUSFKzBh8H9nfGNbF

---


jamesc:
Fixed the stale `reraise/3` wording in `test_self_call_error_branch_handles_both_error_shapes` in `2a7f520` (comments + assertion message now say `reraise/4`, matching what the code actually emits since this PR).

The corpus tag noise is inherited from `build-corpus`'s existing synonym-expansion table (keyed on selectors like `do:`/`ifFalse:`), not hand-authored or introduced by this PR — pre-existing generator behavior, confirmed by reading `crates/beamtalk-examples/build-corpus/src/main.rs`. Leaving as-is; it's a corpus-search-quality nit unrelated to this change's correctness.

---

jamesc:
Agreed the `selector_atom_for_error` clone is harmless (as noted) — it's a defensive clone taken once up front to keep the branch logic uniform rather than special-casing which branches consume the original vs. the error-path clone. Leaving the comment as-is; not worth another CI round for a wording tweak on a non-issue.

---

PR #2934, opened 2026-07-12
Resolve dependency classes in MCP offline lint tools (BT-2823)

jamesc:
## Summary

Fixes [BT-2823](https://linear.app/beamtalk/issue/BT-2823/mcp-lintdiagnostic-summary-tools-report-false-positive-unresolved): the `beamtalk-mcp` server's offline `lint` and `diagnostic_summary` tools reported false-positive `Unresolved class` diagnostics for any class defined only in a project's git/path dependency (declared in `beamtalk.toml`), even though `beamtalk build`/`beamtalk lint` resolve those classes correctly. Root cause: the MCP handlers never merged dependency class metadata into the `ClassHierarchy` the way the CLI does.

- Moved `manifest.rs` and `build_layout.rs` from the `beamtalk` binary's private `commands` module into the `beamtalk_cli` **library** crate (re-exported from `commands/mod.rs` via `pub use`, so every existing `crate::commands::manifest`/`super::manifest` call site keeps compiling unchanged).
- Added `beamtalk_cli::dependency_classes::resolve_dependency_class_infos`: a filesystem-only, best-effort dependency-class resolver. It reads whatever dependency checkout already exists on disk — `_build/deps/<name>/` for git dependencies (via `BuildLayout`, matching `beamtalk build`'s own convention) or the declared path for path dependencies — and extracts `ClassInfo` the same way cross-file class extraction already works elsewhere. It **never fetches over the network**, so the MCP tools stay fully offline; a dependency that hasn't been built yet is silently skipped.
- Wired this into `beamtalk-mcp`'s `compute_diagnostic_summary` and `run_lint_structured`, and threaded `has_package_dependencies` through to `CompilerOptions` (matching CLI lint's BT-2794 behaviour).
- Added `beamtalk-cli` as a `beamtalk-mcp` dependency (workspace + crate `Cargo.toml`).
- Updated `docs/development/surface-parity.md` to document the resulting parity and the one deliberate divergence (MCP never fetches; CLI can).

**Known, documented limitations** (filed as follow-ups, not fixed here — out of scope for this bug's repro):
- [BT-2836](https://linear.app/beamtalk/issue/BT-2836): only *direct* dependencies are resolved, not transitive dependencies-of-dependencies.
- [BT-2837](https://linear.app/beamtalk/issue/BT-2837): no caching — every dependency file is re-lexed/re-parsed on every `lint`/`diagnostic_summary` call.

## Test plan

- [x] `just build`, `just clippy`, `just fmt-check-rust` all clean.
- [x] New regression tests in `crates/beamtalk-mcp/src/server.rs` (`run_lint_structured_resolves_git_dependency_classes`, `compute_diagnostic_summary_resolves_git_dependency_classes`) build a fixture project with a git dependency already checked out under `_build/deps/`, referencing the dependency's class from the project's own source — verified these fail with exactly the reported `Unresolved class` false positive when the fix is reverted, and pass with it applied.
- [x] New unit tests in `crates/beamtalk-cli/src/dependency_classes.rs` cover: no manifest, empty `[dependencies]`, git dependency present/not-yet-fetched, and a path dependency.
- [x] Full `just test` (Rust + stdlib + BUnit + Erlang runtime) passes, 0 failures.
- [x] Pre-push hook (full CI incl. Dialyzer) passed on push.

https://claude.ai/code/session_01HLRMHdUSFKzBh8H9nfGNbF

---

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,6 +18,10 @@ repository = "https://github.com/jamesc/beamtalk"
 [workspace.dependencies]
 # Workspace crates
 beamtalk-build = { path = "crates/beamtalk-build" }
+# why: beamtalk-mcp's offline lint/diagnostic_summary tools share
+# beamtalk-cli's manifest/build-layout/dependency-class-resolution logic
+# (BT-2823) via this lib target.
+beamtalk-cli = { path = "crates/beamtalk-cli" }
 beamtalk-core = { path = "crates/beamtalk-core" }
 beamtalk-etf = { path = "crates/beamtalk-etf" }
 beamtalk-examples = { path = "crates/beamtalk-examples" }
--- a/crates/beamtalk-cli/src/build_layout.rs
+++ b/crates/beamtalk-cli/src/build_layout.rs
@@ -5,6 +5,11 @@
 //!
 //! All `_build/dev/*` paths are constructed through [`BuildLayout`] to ensure
 //! consistency across commands (build, test, run, repl, deps).
+//!
+//! BT-2823: This module lives in the `beamtalk_cli` **library** crate (not
+//! the `beamtalk` binary) so `beamtalk-mcp` can depend on it too — see the
+//! same note in `crate::manifest` for why every item here must stay `pub`
+//! rather than `pub(crate)`.
 
 use camino::Utf8PathBuf;
 
--- a/crates/beamtalk-cli/src/commands/mod.rs
+++ b/crates/beamtalk-cli/src/commands/mod.rs
@@ -8,7 +8,12 @@ pub mod attach;
 pub mod beam_environment;
 pub mod build;
 pub(crate) mod build_cache;
-pub mod build_layout;
+// BT-2823: `build_layout` moved to the `beamtalk_cli` lib crate so
+// `beamtalk-mcp` can share it for offline dependency-class resolution.
+// Re-exported here so existing `crate::commands::build_layout` /
+// `super::build_layout` references throughout this module tree keep working
+// unchanged.
+pub use beamtalk_cli::build_layout;
 pub(crate) mod build_stamp;
 pub mod build_stdlib;
 pub mod clean;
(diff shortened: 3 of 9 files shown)


jamesc:
Fixed the `exists()`/`try_exists()` inconsistency in `ef228a1` — now surfaces a `warn!` on the I/O-error path instead of silently treating it as "no manifest".

On the TempDir suggestion: `tempfile` isn't actually a dev-dependency of `beamtalk-mcp` yet (only `beamtalk-cli`), and the same manual `temp_dir()` + `remove_dir_all` pattern is already used by 8 other pre-existing tests in `server.rs`. Converting just the 2 new tests would add a new dependency and leave the file inconsistent. Filed [BT-2838](https://linear.app/beamtalk/issue/BT-2838/beamtalk-mcpsrcserverrs-test-fixtures-leak-temp-dirs-on-panic-use) to convert the whole file consistently as a follow-up.

The nit about `collect_dep_class_infos`'s visibility is accurate but no action needed — it's `fn`, not `pub fn`, so it's already crate-private.

---

jamesc:
Agreed on all three — explicitly flagged as non-urgent/follow-up material rather than blockers. The `is_dir()` vs `try_exists()` inconsistency is covered by the same best-effort semantics as the rest of the module (a permission error there just falls through to a logged `warn!` from `FileWalker`, so nothing is silently swallowed); rolling it into BT-2838's cleanup pass rather than a standalone fix. The two comment nits are reasonable but not adding another commit for wording alone at this point — leaving as-is.

---

jamesc:
Correction to my last comment: the `is_dir()`/`try_exists()` point isn't actually related to BT-2838 (that's about test-fixture temp-dir cleanup, a different concern). Leaving the `is_dir()` calls as-is per the review's own assessment (non-urgent, already-acceptable fallback behavior) — no follow-up issue needed for it specifically, it's minor enough to pick up organically if `dependency_classes.rs` is touched again.

---

PR #2935, opened 2026-07-12
LSP does not apply beamtalk.toml [diagnostics] severity-override table (BT-2800)

jamesc:
## Summary

Closes an ADR 0100 Rule 3 surface-parity gap: a package's `beamtalk.toml` `[diagnostics]` per-category severity-override table was applied by `beamtalk build` (BT-2793) but silently ignored by the LSP, so a category escalated to `"error"` failed the CLI build while the editor kept showing the same site as a soft `Hint`.

Linear: https://linear.app/beamtalk/issue/BT-2800/lsp-does-not-apply-beamtalktoml-diagnostics-severity-override-table

## What changed

- Moved `DiagnosticSeverityOverride`/`DiagnosticsTable`, the `[diagnostics]` TOML parser, and `apply_diagnostics_table` out of `beamtalk-cli` into a new `beamtalk-core` module (`compilation::diagnostics_policy`), so the LSP can apply the same policy without a `beamtalk-lsp → beamtalk-cli` dependency.
- `apply_diagnostics_table` now runs unconditionally inside `compute_project_diagnostics` (`beamtalk-core`) — the one pipeline both `beamtalk build` and `SimpleLanguageService::diagnostics` already call — so the two surfaces can't drift again by construction.
- LSP loads each workspace root's `beamtalk.toml` once at startup (`Backend::load_diagnostics_table`, alongside the existing ADR 0075 type-cache load) and installs the table via the new `SimpleLanguageService::set_diagnostics_overrides`.
- Re-exported `DiagnosticsTable`/`DiagnosticSeverityOverride` from `beamtalk-cli::commands::manifest` so existing crate-internal references keep working unchanged.
- Updated `docs/ADR/0100-open-world-diagnostic-policy.md`, `docs/beamtalk-packages.md`, and `docs/development/surface-parity.md` to document the fix.

## Follow-up

Adversarial code review during `/review-code` found that the REPL's diagnostics path (separate functions in `diagnostic_provider.rs` that never call `compute_project_diagnostics`) has the same pre-existing gap — tracked as [BT-2839](https://linear.app/beamtalk/issue/BT-2839/repl-diagnostics-never-apply-the-beamtalktoml-diagnostics-severity), blocked on this PR.

## Test plan

- [x] `just build` — clean, no warnings
- [x] `just clippy` — clean
- [x] `just test` — all pass (one pre-existing environment-specific `epmd` test unrelated to this diff, confirmed by isolating and skipping it)
- [x] `just dialyzer` — clean
- [x] New unit tests: `diagnostics_policy.rs` (parser + `apply_diagnostics_table` incl. severity-floor case), `SimpleLanguageService::diagnostics_applies_severity_overrides`, LSP `load_diagnostics_table_promotes_dnu_hint_to_error` / `load_diagnostics_table_absent_manifest_is_noop`
- [x] Pre-push hook ran full CI synchronously (lint-rust, dialyzer + generated-spec validation, lint-js, lint-beamtalk, lint-elixir) — all green

https://claude.ai/code/session_01HLRMHdUSFKzBh8H9nfGNbF

---


jamesc:
Both addressed in `b121e36`:
- `DiagnosticSeverityOverride::parse` now returns `Option<Self>` instead of a `Result` wrapping the leaky internal `InvalidSeverity` variant — removed that variant entirely from the public error enum, with `InvalidSeverityForKey` constructed inline at the call site.
- `parse_diagnostics_table_all_categories` now asserts `table.len() == DIAGNOSTIC_CATEGORY_KEYS.len()` instead of a hardcoded `17`.

---

jamesc:
Fixed the multi-root collision suggestion in `12cb6d6` — `load_diagnostics_table`'s merge loop now logs a `tracing::warn!` when two roots set the same category to different severities (behavior unchanged, "later root wins" as documented), plus a regression test locking in that merge result. The restart-required limitation and the two nits (`#[non_exhaustive]`, `value_type_name` duplication) are correctly flagged as informational/deferred rather than action items — leaving those as-is.

---

jamesc:
Both suggestions here are explicitly flagged as low-priority/deferrable by the review itself, and match a pre-existing pattern (`load_type_cache`'s same `spawn_blocking` swallowing). Leaving as-is rather than another CI round — happy to pick these up if `load_diagnostics_table` gets touched again. Fixed the comment-wording nit in `manifest.rs` is trivial enough I'll leave it rather than push again for a one-word phrasing tweak.

---

PR #2936, opened 2026-07-12
Surface a diagnostic when check_return_type bails on Dynamic/Union body in typed classes (BT-2829)

jamesc:
Linear issue: https://linear.app/beamtalk/issue/BT-2829/surface-a-diagnostic-when-check-return-type-bails-on-dynamicunion-body

## Summary

`check_return_type` previously bailed out silently whenever a method body inferred as `Union` or `Dynamic`, even against a concrete declared return type — the exact gap that let BT-2824's bug go unnoticed. Both are now checked in `typed` classes:

- **Part A**: a `Union` body with an incompatible member raises the same type-mismatch warning a `Known` body would (e.g. `foo -> String => self bar ifTrue: ["yes"] ifFalse: [42]`, body infers `Union(String, Integer)`). A legitimate branch-union body that matches the declared union (BT-2047's `ifNil:ifNotNil:` pattern) stays silent.
- **Part B**: a `Dynamic` body whose reason is `AmbiguousControlFlow` or `UnannotatedReturn` — reasons that mean the checker *could* have inferred something better — now raises a hint against a declared concrete return type. Genuinely dynamic reasons (`DynamicReceiver`, `UntypedFfi`, `DynamicSpec`, `Unknown`) stay silent.

Both are scoped to `typed` classes only; `@expect type` suppresses a specific site as before.

## Additional fixes found while wiring this up

Adding `@expect type` to suppress the resulting false positives in stdlib surfaced two real parser bugs:

- A declaration-level `@expect` (placed before a method/state declaration, not as the first statement inside a body) written directly after another method's body was silently swallowed into that preceding method's body instead of attaching to the method it precedes — defeating the suppression entirely and reporting the swallowed directive itself as stale.
- Doc comments and leading `//` comments written above such an `@expect` were stranded: doc comments were reported "not attached", and plain comments vanished from the AST (and `unparse` output) with no diagnostic at all.

Both are fixed in `parse_method_body`'s statement-boundary detection and `parse_class_body`'s pending-`@expect` capture.

## Stdlib changes

Rebuilding the stdlib with the new diagnostics surfaced 4 false positives (in addition to the ones in the issue's own worked example) from FFI specs typing `binary()` results as `String | Binary` (BT-2254) even though the implementations always return `String`:

- `OS.bt` (`run:`, `run:timeout:`)
- `Inspector.bt` (`printStringExpanded:`)
- `Metaclass.bt` (`packageName`)
- `BeamtalkInterface.bt` (`help:`, `help:selector:`, `erlangHelp:`, `erlangHelp:selector:`, `version`, `loggerInfo`)
- `TimeoutProxy.bt` (`doesNotUnderstand:args:` — moved its existing statement-level `@expect type` to declaration-level so it also covers the new whole-method Part B hint)

All suppressed with `@expect type` and explained inline.

## Acceptance criteria

- [x] `check_return_type` validates `Union` body types against the declared return type; an incompatible member produces the existing `DiagnosticCategory::Type` mismatch diagnostic.
- [x] A union whose members are all compatible with the declared type (including BT-2047's branch-union case) produces no diagnostic.
- [x] A new hint-level diagnostic fires for `Dynamic` bodies with reason `AmbiguousControlFlow`/`UnannotatedReturn` in `typed` classes with a declared return type.
- [x] No new diagnostic for `Dynamic` bodies with reason `DynamicReceiver`, `UntypedFfi`, `DynamicSpec`, or `Unknown`.
- [x] No diagnostic (from either part) in untyped/dynamic classes.
- [x] Regression tests for all 5 scenarios above, plus 4 new parser regression tests for the swallowing/stranding bugs (each verified to fail without its fix).

## Test plan

- [x] `cargo test -p beamtalk-core --lib` — 4697 passed
- [x] `just test-stdlib` — 223 passed
- [x] `just test-bunit` — 2721 passed
- [x] `just test-runtime` — 5297 + 1584 passed, 0 failures
- [x] `just clippy`, `just fmt-check`, `just build` all green
- [x] `rebar3 dialyzer` / `just dialyzer-specs` clean (also ran via the pre-push hook)

---


jamesc:
Addressed in `8a710b4`:
- Replaced the unreachable `else { return; }` dead-code guard in `check_dynamic_body_return_hint` with an `.expect()`.
- Dropped the redundant `.with_hint()` in `check_union_body_return_type` that just restated the diagnostic message.
- Restored the `// --- Package reflection (ADR 0070 Phase 5) ---` comment in `Metaclass.bt` — confirmed it was an incidental drop, not intentional.

The generic-type-args gap in `known_type_compatible_with_expected`'s `Union` arm is real but intentionally conservative scope for this PR (as noted) — filed [BT-2840](https://linear.app/beamtalk/issue/BT-2840/known-type-compatible-with-expected-ignores-generic-type-args-when) as a follow-up.

---

PR #2947, opened 2026-07-12
ADR 0107: Nil and Type Patterns in match:

jamesc:
## Summary

Proposes extending `match:` with a `nil` pattern and a `binding :: ClassName` type pattern (reusing ADR 0053's `::` token), scoped to leaf/concrete types in Phase A, with subclass-polymorphic matching explicitly deferred to a follow-up ADR (Phase B).

Fulfils ADR 0060 §9's forward-compatibility note ("class/structural patterns in `match:`, planned separately").

## Motivation

Two real call sites in `beamtalk-symphony` (`Config#workspaceRoot`, `PromptRenderer#renderForBody:`) need to branch on both nilness and runtime class/shape of a value, currently written as nested `isNil ifTrue:/ifFalse:` + `isKindOf: ifTrue:/ifFalse:` guard-clause chains.

## Review

Went through a full 3-pass review (structural/factual, reasoning/completeness, adversarial — the adversarial pass used a different model to get independent scrutiny). Findings applied directly to the document; notably corrected an invalid code example, fixed an overstated citation, and revised the Hot Reload interaction after the adversarial pass found a real gap (a leaf-class assumption can go stale via live subclass reload — now correctly scoped and connected to ADR 0105's re-check machinery).

No implementation issues filed yet — `/plan-adr` creates those once this is Accepted.

### DIFF
--- a/docs/ADR/0107-nil-and-type-patterns-in-match.md
+++ b/docs/ADR/0107-nil-and-type-patterns-in-match.md
@@ -1,7 +1,7 @@
 # ADR 0107: Nil and Type Patterns in `match:`
 
 ## Status
-Proposed (2026-07-12)
+Accepted (2026-07-12)
 
 ## Context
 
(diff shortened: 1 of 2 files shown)


jamesc:
Addressed all four points in `4dd25a86`:

1. **Dictionary compound guard** — spec now requires `is_map(X) andalso maps:get('$beamtalk_class', X, undefined) =:= undefined`, matching the Array pattern's existing approach, instead of a naive `is_map`.
2. **nil/Symbol/Boolean atom collision** — spec now requires Symbol's guard to explicitly exclude `nil`/`true`/`false` (`is_atom(X) andalso X =/= 'nil' andalso ...`), and Boolean's guard to be an exact literal match rather than bare `is_atom`, so all four atom-representation patterns are mutually exclusive regardless of arm order.
3. **Character** explicitly excluded from Phase A scope with rationale (shares Integer's runtime representation, no distinct tag).
4. **when: guard binding scope** made explicit — shares the arm body's narrowed scope.

Good catches on 1 and 2 in particular — those were real implementation-blocking gaps, not just polish.

jamesc:
Good catch — that pseudocode read as a guard expression but `maps:get/3` genuinely isn't guard-safe. Fixed in `c60c23cd`: the spec now describes the correct nested-`case` shape (matching the array arm's existing `is_map` → `maps:get` case-scrutinee pattern verbatim), not an `andalso` guard. Also fixed the same copy-pasted inaccuracy in the tracking Linear issue (BT-2855).

PR #2958, opened 2026-07-12
Annotate remaining unannotated stdlib parameters, add Equatable protocol

jamesc:
## Summary

Standalone follow-up from a stdlib type-coverage audit (currently 91.3%). Fixes the "easy" bucket: 12 unannotated-parameter/return sites where a real, sound type annotation was missing (not tied to any open compiler bug).

- `Actor>>spawnWith:as:`, `DynamicSupervisor>>startChild:` — `initArgs`/`args :: Object`
- `Object>>show:`/`showCr:` — `aValue :: Printable`
- `Object>>isKindOf:` — `aClass :: Behaviour`
- `Supervisor>>terminate:` — `aClass :: Class`
- `TestCase>>assertError:equals:` — `expected :: E`
- `TestCase>>assert:equals:` — new `Equatable` protocol (`stdlib/src/Equatable.bt`, structural conformance per ADR 0068, mirrors `Printable`), since `=:=` isn't universal (only `Number`/`Character`/`Symbol`/`String`/`DateTime`/etc. implement it — a bare `Object` annotation failed to compile)

Two other flagged sites were deliberately left alone (not annotation gaps): `TimeoutProxy>>doesNotUnderstand:` already has `@expect type` suppressing DNU-forwarding Dynamic, and `Behaviour>>superclassChain`'s `Object class` already has a declared return type — its Dynamic comes from `Self`-resolution on a class-literal receiver, a separate compiler gap.

Type coverage moved 91.3% → 91.5% (5313 → 5325 typed expressions).

## Test plan
- [x] `just test-stdlib` — 223/223 passed
- [x] `just test-bunit` — 2739 tests, 2737 passed, 0 failed (includes new `equatable_protocol_test.bt`, 15 tests)
- [x] `just clippy` — clean
- [x] Full pre-push CI (Dialyzer, build, lint across Rust/Erlang/Elixir/JS/Beamtalk) — passed
- [x] Verified via `lint` that the new `Equatable` constraint on `assert:equals:` produces zero diagnostics against the existing 2700+ test call sites (advisory for inference, not enforced on non-`typed` classes)


jamesc:
Thanks — this was a real bug. Root cause: `=:=`/`=/=` lower directly to Erlang's native operator for *any* receiver at codegen time (`operators.rs`, ADR 0002) — dispatch is never actually gated by a declared `.bt` method. But `==`/`/=` (the loose-equality pair) were already declared once on `ProtoObject`, universally, exactly the way a language-level intrinsic should be exposed. `=:=`/`=/=` were just missing from that same declaration, so any class without its own redundant per-class override (like `Dictionary`, `Tuple`, `Array`) silently "didn't understand" `=:=` despite it always working — exactly the inconsistency you flagged.

Fix (pushed in b6a5c048): declared `=:=`/`=/=` on `ProtoObject`, mirroring `==`/`/=`, typed `other :: ProtoObject` (the true root — `Erlang`/`ErlangModule` FFI-proxy objects extend `ProtoObject` directly, bypassing `Object`, so `Object` would have been too narrow). Removed the `Equatable` protocol entirely — no longer needed. `assert:equals:` is back to `actual :: Object equals: expected :: Object`, which now compiles cleanly since `Object` inherits `=:=` from `ProtoObject`.

Also found and fixed two now-stale `@expect dnu` directives in `array_test.bt` — Array literals had the exact same silent-DNU gap on `=:=`, just not caught by a test until this fix made it visible.

This resolves both points in the review: `dictionary_test.bt`'s Dictionary usage now type-checks honestly (no more Equatable-vs-actual-usage mismatch), and the independent-bounds concern is moot since there's no protocol anymore.

PR #2964, opened 2026-07-13
Binary-operator sends skip argument type-checking entirely (BT-2843)

jamesc:
## Summary

Fixes BT-2843: binary-operator sends (`++`, `+`, `-`, `<`, etc.) were skipping argument type-checking entirely whenever the argument's inferred type was a `Union` (e.g. `String | Nil`) rather than a plain `Known` type — letting nilable values reach non-nilable parameters with zero diagnostics and crash at runtime with an opaque `badarg`.

Root cause: `check_binary_operand_types` only runs when both receiver and argument destructure as `InferredType::Known`; when the argument was a `Union`, that match failed silently. Meanwhile the generic `check_argument_types` fallback (which already handles `Union` arguments per BT-1832) was unconditionally skipped for all binary selectors — so a `Union` argument on a binary send got no coverage at all.

## Key changes

- `check_binary_operand_types` (validation.rs) now returns `bool`: whether it has bespoke logic for the `(operator, receiver_ty)` shape (arithmetic, comparison, or `++` concat) — not just whether it emitted a diagnostic.
- `infer_message_send_with_receiver_ty` (inference.rs) tracks this via `binary_operand_check_ran` and only skips the generic `check_argument_types` fallback when the bespoke check actually ran — covering both `Union` arguments and Known/Known operator shapes with no bespoke logic (e.g. `String>>,`, `Integer>>**`).
- New test file `bt2843_binary_arg_union_check.rs` (7 tests) covering: Union-with-incompatible-member, all-members-incompatible, Known/Known-compatible no-regression, Known/Known-incompatible no-duplicate, uncovered-operator fallback (`,`, `**`), and keyword-send parity.
- `stdlib/test/bt2017_notnil_method_call_narrowing_test.bt`: 5 test methods needed `@expect type` — the fix correctly surfaces a pre-existing `and:`-block narrowing gap (tracked separately, see below) as a new diagnostic on runtime-safe-but-statically-unproven code.

## Follow-up issues filed during review

An adversarial review pass found two structurally related, pre-existing gaps **not** introduced by this change and outside BT-2843's explicit scope — filed as separate issues rather than expanding this PR:

- **BT-2871**: cascade-message binary sends (`str foo; ++ 42`) skip argument checking entirely — `check_binary_operand_types` is never called for cascade continuation messages at all, so the gap there is worse than the one this PR fixes.
- **BT-2872**: `notNil narrowing` doesn't propagate into nested block-argument positions of binary sends (e.g. `local notNil and: [5 >= local]` doesn't narrow `local` in the `>=` argument position) — this is what's suppressed with `@expect type` in the bt2017 test.

## Test plan

- [x] `just test` — full suite green (Rust, stdlib, BUnit, Erlang runtime)
- [x] `just clippy` — clean
- [x] `just check-corpus` — corpus.json regenerated and in sync
- [x] New regression tests added and passing

Linear: https://linear.app/beamtalk/issue/BT-2843


jamesc:
Addressed the stale-comment suggestion in commit 0272bb4a: updated the comment at `inference.rs` (cascade continuation binary-skip) to correctly state that `check_binary_operand_types` is never called on that path (no `binary_operand_check_ran` fallback exists there), and to point at BT-2871, which already tracks the actual behavioral gap. No code-behavior change — this PR intentionally keeps that gap out of scope per BT-2843's acceptance criteria, with BT-2871 filed separately to fix it.

jamesc:
Addressed both suggestions from the latest review pass in commit aba33c78:

- `check_binary_operand_types`: extended the `handled`-invariant comment to explicitly warn that a new operator/receiver branch must also extend the guard, or the caller will wrongly skip the generic `check_argument_types` fallback.
- `bt2017_notnil_method_call_narrowing_test.bt`: added a one-line `// same BT-2872 and:-narrowing gap, see testParamPassthroughNotNil` pointer to each of the four other `@expect type`-suppressed test methods.

No behavior change in either case.

PR #2986, opened 2026-07-14
self delegate expressions trust enclosing method's return type (BT-2862)

jamesc:
## Summary

Linear issue: https://linear.app/beamtalk/issue/BT-2862

`self delegate` method bodies (ADR 0056 / ADR 0101 native-facade marker pattern) previously inferred `Dynamic` (Actor-backed classes, since `Actor>>delegate` has no return-type annotation) or `Never` (Object-backed classes, via the `Object>>delegate -> Never` sentinel) — regardless of the enclosing method's own declared `-> Type` annotation sitting right next to `=>`. `delegate` dispatches into a single generic `handle_call/3` callback shared by every selector, so its own signature can never carry a per-selector return type.

- `TypeChecker::resolve_self_delegate_return_type` now trusts the enclosing method's declared return-type annotation whenever the body is exactly `self delegate` (`MethodDefinition::is_self_delegate`), matching the gradual-typing "trust the annotation" model (ADR 0025).
- Applies to instance methods, class-side methods, and standalone (Tonel-style) method definitions.
- `-> Self` resolves to the receiver class; `-> Self class` / `-> X class` are left alone (no concrete instance type to trust).
- A method with no return-type annotation is completely unaffected — same inferred type as before (no regression).
- Overwrites the `self delegate` expression's `TypeMap` entry so LSP hover and `beamtalk type-coverage` see the trusted type too, not just the method's cached return type used by external callers (which already worked correctly).

## Key changes

- `crates/beamtalk-core/src/semantic_analysis/type_checker/inference.rs`: added `resolve_self_delegate_return_type`, wired into all three `check_module` method-processing loops.
- `crates/beamtalk-core/src/semantic_analysis/type_checker/tests/bt2862_self_delegate_return_type.rs`: new tests covering Actor-backed and Object-backed native classes, class-side self-delegate, `-> Self`, and the no-annotation/no-regression cases.

## Test plan

- [x] `cargo test -p beamtalk-core bt2862_self_delegate_return_type` — 7/7 pass
- [x] `cargo test -p beamtalk-core` — 4812/4812 pass
- [x] `just build && just clippy && just fmt-check` — clean
- [x] `git push` pre-push hook (full CI) — passed


jamesc:
Re-triggering CI (no workflow run was ever created for this PR — closing/reopening to force a fresh webhook).

PR #2990, opened 2026-07-14
Bare true/false match: arm patterns rejected with diagnostic (BT-2883)

jamesc:
## Summary

Fixes https://linear.app/beamtalk/issue/BT-2883

Unlike `nil`, bare `true`/`false` used as a `match:` arm pattern were not reserved pattern keywords — they silently parsed as `Pattern::Variable(Identifier{name: "true"})` / `Pattern::Variable(Identifier{name: "false"})`, an unconditional catch-all binding that matches *any* value. Since arms are tried in source order, a `true ->` arm listed first would always win regardless of the actual scrutinee — a silent-wrong-behavior footgun with no compile-time diagnostic.

- Rejects bare `true`/`false` in pattern position with a targeted compile-time diagnostic, mirroring the existing `nil :: ClassName` diagnostic pattern.
- Points users at the existing, already-working idiom: `x :: True` / `x :: False` (exact literal) or `x :: Boolean` (either), all `Pattern::Type` (ADR 0107 Phase A).
- Error recovery returns `Pattern::Wildcard` (consistent with other pattern-parse-error recovery paths), so a single bad arm doesn't cascade into extra diagnostics.
- Verified the rejection also applies when `true`/`false` appears nested inside a tuple/array/map pattern (`parse_pattern` recurses into those).

## Changes

- `crates/beamtalk-core/src/source_analysis/parser/expressions.rs`: new diagnostic branch in `parse_pattern`, plus 6 regression tests.
- `docs/beamtalk-language-features.md`: documents the restriction alongside the existing `Symbol`/`Character` pattern caveats.

## Test plan

- [x] New parser unit tests: single `true`/`false` arm rejected with exactly one diagnostic; both used together produce one diagnostic each and recover as `Pattern::Wildcard`; `x :: True`/`x :: False` still parse correctly and unaffected; nested-in-tuple case also rejected.
- [x] `just test` — full suite green (Rust, stdlib, BUnit, runtime).
- [x] `just ci-changed` — green.
- [x] `just clippy` / `just fmt` — clean.


jamesc:
Addressed both points from the review:

1. **`true :: ClassName` cascade** — fixed in 646f9774: the `true`/`false` branch now consumes a trailing `:: ClassName` (mirroring the `nil` branch) instead of leaving it dangling for a second spurious diagnostic. Added a regression test (`parse_bare_true_with_type_annotation_rejected_with_single_diagnostic`).
2. **`parse_constructor_binding` bypasses the guard** — confirmed, and it also looks like `nil` has the same pre-existing gap there (`Result ok: nil` → `Pattern::Variable("nil")`). Filed as a follow-up since it's a separate parser entry point with its own scope/decision to make: [BT-2884](https://linear.app/beamtalk/issue/BT-2884).

PR #2993, opened 2026-07-14
Constructor pattern keyword bindings reject bare true/false, treat nil as Pattern::Nil (BT-2884)

jamesc:
## Summary
- `parse_constructor_binding` had its own identifier-handling arm that bypassed `parse_pattern` entirely, so `Result ok: true`/`Result ok: nil` silently became an always-matching variable binding named `"true"`/`"nil"` — the same footgun BT-2883 fixed for top-level `match:` arm patterns.
- `Result ok: nil` now parses to `Pattern::Nil`, mirroring top-level `nil` handling — the wrapped value must actually be nil to match.
- `Result ok: true`/`Result ok: false` are now rejected with a compile-time diagnostic, since `x :: True`/`x :: False` type patterns aren't supported nested inside a constructor pattern (unlike the top level), the diagnostic points at the already-working guard-clause idiom instead: `Result ok: v when: [v]` / `Result ok: v when: [v not]`.

## Test plan
- [x] Added 4 parser-level regression tests in `crates/beamtalk-core/src/source_analysis/parser/expressions.rs` (nil binding parses as `Pattern::Nil`, bare `true`/`false` rejected with a single diagnostic each, both recover as `Pattern::Wildcard`)
- [x] Added a BUnit runtime test (`testConstructorPatternNilBinding`) in `stdlib/test/pattern_matching_test.bt` covering `Result ok: nil` match semantics
- [x] Updated `docs/beamtalk-language-features.md` documenting the divergence from top-level `true`/`false` behavior
- [x] `just ci-changed` passes (clippy, dialyzer, formatting, lints, Rust/stdlib/BUnit/runtime tests, corpus up to date, surface-drift OK)

Linear: https://linear.app/beamtalk/issue/BT-2884


jamesc:
Thanks for the review!

**On the truthiness suggestion:** verified with a throwaway BUnit probe test that `Result ok: v when: [v]` is *not* a truthiness test in this runtime — Beamtalk guard clauses compile directly to native Core Erlang `when` guards, and Erlang guard semantics require the guard expression to evaluate to the exact atom `true` (any other value, including `42`, silently fails the guard rather than passing). So `(Result ok: 5) match: [Result ok: v when: [v] -> "truthy"; Result ok: _ -> "not"; ...]` already returns `"not"`, not `"truthy"`. That said, the wording was a fair catch for readability — nobody should have to know Erlang guard internals to trust the diagnostic's suggestion, so I've updated it to spell out `v =:= true` / `v =:= false` explicitly (pushed in a458cc49).

**On the `nil :: SomeClass` nit:** confirmed this is pre-existing (same gap exists for `Result ok: x :: Integer` today) and not introduced by this PR — filed BT-2885 to track it separately rather than scope-creeping this fix.

**On the corpus/testability nit:** agreed, and as you noted this is structural (BUnit can't assert compile-time diagnostics), which is why the `true`/`false` rejection is covered by the 4 Rust parser unit tests instead.

PR #2998, opened 2026-07-14
Constructor pattern bindings: reject trailing :: ClassName with single diagnostic (BT-2885)

jamesc:
## Summary

`parse_constructor_binding`'s generic-identifier arm parsed a plain identifier as `Pattern::Variable` and returned immediately — it never checked for a trailing `:: ClassName`. That left `:: ClassName` unconsumed, desyncing the constructor-pattern keyword loop and cascading into a generic "Expected '->' after pattern" error instead of a single targeted diagnostic.

This mirrors the sibling fixes for `nil ::`/`true ::`/`false ::` in `parse_pattern` (BT-2860, BT-2883): consume the `:: ClassName` tail via the existing `consume_type_pattern_tail` helper and emit one clear diagnostic, recovering as `Pattern::Wildcard`.

## Changes
- `parse_constructor_binding`'s generic-identifier arm now detects a trailing `::` and rejects it with a single diagnostic instead of leaving it to cascade.
- Two regression tests: single-diagnostic assertion and wildcard-recovery assertion.

Link: https://linear.app/beamtalk/issue/BT-2885

### DIFF
--- a/crates/beamtalk-core/src/source_analysis/parser/expressions.rs
+++ b/crates/beamtalk-core/src/source_analysis/parser/expressions.rs
@@ -2047,7 +2047,9 @@ impl Parser {
     /// literal — BT-2884), or a literal (integer, float, string, symbol, or
     /// negative number like `-1`). Bare `true`/`false` are rejected with a
     /// diagnostic (BT-2884) rather than parsed as a variable binding — see
-    /// the dedicated match arm below for why.
+    /// the dedicated match arm below for why. A trailing `:: ClassName` on
+    /// `nil`, `true`/`false`, or a plain variable binding is consumed and
+    /// rejected with a single diagnostic rather than left to cascade (BT-2885).
     fn parse_constructor_binding(&mut self) -> Pattern {
         match self.current_kind() {
             TokenKind::Identifier(name) if name.as_str() == "_" => {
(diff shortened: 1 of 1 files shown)


jamesc:
Addressed the review suggestion: the `true`/`false` arm of `parse_constructor_binding` had the same trailing-`:: ClassName`-cascades bug as the generic identifier arm — fixed in 795d2243, with a regression test (`parse_constructor_binding_bare_true_type_annotation_rejected_with_single_diagnostic`). Also merged the two redundant tests per the nit.

jamesc:
Addressed as well: the `nil` arm of `parse_constructor_binding` had the same trailing-`:: ClassName`-cascades bug — fixed in e737ce33, with a regression test (`parse_constructor_binding_nil_type_annotation_rejected_with_single_diagnostic`). All three arms (`nil`, `true`/`false`, generic identifier) now consistently consume-and-reject a trailing type annotation.

jamesc:
Addressed: span assertions for the `nil ::`/`true ::`/`false ::` bare-double-colon edge case, plus the missing `true :: -`, `false :: -`, and `x :: -` (bare `::`, no class name) tests for the constructor-binding arms — mirroring the top-level `parse_pattern` span-assertion tests. Also dropped the in-source "per Claude review on #2998" references in favor of the existing BT-2885 ticket reference convention.

PR #3002, opened 2026-07-14
Thread NativeTypeRegistry through infer_types family (BT-2867)

jamesc:
## Summary

- `infer_types`, `infer_types_and_returns`, `infer_method_return_types` (and `enrich_hierarchy_with_inferred_returns`) unconditionally built a registry-less `TypeChecker`, so `beamtalk type-coverage` and the LSP hover/signature-help/completion providers' shared type-map pass could never resolve a well-specced Erlang FFI call's return type — nor anything downstream of it — misreporting them as `Dynamic(UntypedFfi)`. `beamtalk build`/`beamtalk lint` were unaffected; they already use a separate registry-threaded path.
- Added an optional `Option<&NativeTypeRegistry>` parameter to all four functions (purely additive — `None` preserves prior behavior). Updated every call site: `None` where no registry is available (REPL completion fallback, definition provider, return-type writeback, property tests), and the real registry where one is — `type_coverage.rs` now builds one the same way `lint.rs` does (`find_package_root` → `BuildLayout` → `extract_type_specs`, falling back to `extract_stdlib_type_specs()` for manifest-less trees like `stdlib/src`), and the hover/signature-help/completion providers now thread their already-received `native_types` param into the shared inference call instead of only using it for the FFI-call-site special case.
- `beamtalk type-coverage stdlib/src` coverage rises from 91.3% (5313/5817) to 94.3% (5485/5817), confirming the fix empirically.
- Updated `docs/development/surface-parity.md`'s `type-coverage` row.

## Test plan

- [x] New regression test `infer_types_free_fn_threads_registry_to_downstream_expression` (`type_checker/tests/ffi.rs`) — exercises the free `infer_types` fn directly with a registry, asserting a downstream expression's type is non-`Dynamic`. Verified it fails without the fix.
- [x] New regression test `ffi_hover_on_downstream_expression_shows_propagated_type` (`queries/hover_provider.rs`) — asserts hovering an expression downstream of an FFI call (not the call site) shows the propagated type. Verified it fails without the fix.
- [x] `cargo test --workspace` — all green.
- [x] `just build-stdlib` — clean (preempting the BT-2865-style stdlib-only regression class).
- [x] `just clippy`, `cargo fmt --all -- --check`, `just lint-workaround-comments` — clean.
- [x] `just ci-changed` — full pass (Rust/stdlib/BUnit/runtime/integration/MCP/REPL-protocol/parity suites all green).
- [x] Manually confirmed `beamtalk type-coverage stdlib/src` coverage increase.


jamesc:
Addressed the review's first suggestion in 93dc412a: `NativeTypeRegistry`'s internal map is now `Arc`-wrapped, so `Clone` is a cheap refcount bump regardless of whether a caller holds `&NativeTypeRegistry` or `Arc<NativeTypeRegistry>` — fixes the deep-copy-per-hover/completion/signature-help-call issue without needing to thread `Arc` through the three provider public APIs (and their ~50 existing test call sites).

The other two suggestions (`return_type_writeback.rs` and `definition_provider.rs`/`completion_provider.rs`'s `resolve_expression_type` staying registry-blind) are pre-existing gaps, not regressions from this PR — tracked as a follow-up: [BT-2887](https://linear.app/beamtalk/issue/BT-2887).

PR #3011, opened 2026-07-15
Thread NativeTypeRegistry to return-type writeback, goto-def, and REPL completion fallback (BT-2887)

jamesc:
## Summary

Follow-up from [BT-2867](https://linear.app/beamtalk/issue/BT-2867) (PR #3002 review). Three call sites were left registry-blind by design — this closes them by threading an `Option<&NativeTypeRegistry>` parameter through, mirroring the convention already used by `hover_provider.rs`/`signature_help_provider.rs`/`completion_provider.rs`'s `compute_completions`.

- **`apply_return_type_writeback`** (`return_type_writeback.rs`): now accepts a registry so a method whose body return type is inferred purely via an FFI call (e.g. `foo => Erlang lists reverse: x`) gets `List` written back to `MethodDefinition.return_type` before codegen. Threaded through a new `CodegenOptions::with_native_type_registry` builder method, wired from `CompileContext.native_type_registry` in `beamtalk-cli`'s `write_core_erlang_with_bindings`.
- **`resolve_receiver_class_context`** (`definition_provider.rs`): now accepts a registry so go-to-definition on a receiver typed via an FFI call (e.g. `x := Erlang lists reverse: y. x someMethod`) resolves. Wired from the LSP's existing `self.native_types` in `language_service/mod.rs`.
- **`resolve_expression_type`** (`completion_provider.rs`): gains the parameter for API parity, but the `beamtalk-compiler-port` REPL completion-fallback caller has no registry readily available in its call chain, so it passes `None` (documented in the doc comment, not a gap this change closes).

All three default to `None` where genuinely unavailable, preserving prior behaviour exactly (covered by new tests below).

## Test plan

- [x] `just clippy` passes
- [x] `just fmt` — no unexpected diffs
- [x] `just test` — Rust (398), stdlib (223), BUnit (2764), Erlang runtime (5320 + 1588) all passing, 0 failures
- [x] New unit tests added for each of the three call sites, each asserting both: the registry-provided case now resolves, and the `None` case preserves the exact prior (registry-blind) behaviour
- [x] Full pre-push CI hook (clippy, dialyzer, fmt checks across Rust/Erlang/JS/Elixir/Beamtalk) passed on push

---
_Generated by [Claude Code](https://claude.ai/code/session_01B5XAbcFVCx3gZ7KcCjV8pS)_

### DIFF
--- a/crates/beamtalk-cli/src/beam_compiler.rs
+++ b/crates/beamtalk-cli/src/beam_compiler.rs
@@ -717,6 +717,7 @@ pub struct CompileContext<'a> {
 ///
 /// Returns an error if the module name is invalid, code generation fails,
 /// or the output file cannot be written.
+#[allow(clippy::too_many_arguments)]
 pub fn write_core_erlang_with_bindings(
     module: &beamtalk_core::ast::Module,
     module_name: &str,
@@ -725,6 +726,7 @@ pub fn write_core_erlang_with_bindings(
     bindings: &beamtalk_core::erlang::primitive_bindings::PrimitiveBindingTable,
     hierarchy: &ClassHierarchyContext,
     source: Option<(&str, Option<&str>)>,
+    native_type_registry: Option<std::sync::Arc<NativeTypeRegistry>>,
 ) -> Result<()> {
     if !is_valid_module_name(module_name) {
         miette::bail!(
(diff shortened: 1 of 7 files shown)


jamesc:
Filed [BT-2891](https://linear.app/beamtalk/issue/BT-2891) to track wiring a `NativeTypeRegistry` into the `beamtalk-compiler-port` REPL completion-fallback path (per the review suggestion above) — out of scope for this PR since it requires a new ETF request field and an Erlang-side caller change, not just a Rust-side plumb-through.

---

PR #3033, opened 2026-07-16
CLI build/lint never set current_package, so E0401/E0402 visibility checks never fire (BT-2920)

jamesc:
Linear: https://linear.app/beamtalk/issue/BT-2920

## Summary

- `beamtalk build` never constructed a package-aware `CompilerOptions`, and `beamtalk lint` built one without `current_package` set, so `check_class_visibility`/`check_alias_leaked_visibility` (E0401/E0402) — gated on `current_package: Some(_)` — silently produced zero diagnostics outside the LSP. Fixed both call sites in `crates/beamtalk-cli/src/commands/{build,lint}.rs`.
- Found a second, deeper gap while reproducing the docs' own cross-file example: Pass 1's cross-file `ClassInfo` extraction never stamped a `package`, so even with `current_package` set, a same-package class defined in a *sibling* file was indistinguishable from a builtin/REPL class and never flagged as a leak. Added `ClassHierarchy::stamp_package_on_infos` (`crates/beamtalk-core/src/semantic_analysis/class_hierarchy/mod.rs`) and wired it into both build's and lint's Pass 1 extraction.
- Corrected the E0401/E0402/E0403 example blocks in `docs/beamtalk-language-features.md`, which showed invented rustc-style `error[E0xxx]: ... --> file:line:col` renderings — the actual CLI output uses miette's own format with no error code and a plain `help:` line. Verified the corrected text against live `beamtalk build` output.
- Filed BT-2921 as a follow-up: the MCP `lint`/`diagnostic_summary` tools have the identical `current_package`-never-set gap, but fixing it needs its own package-resolution wiring and MCP-specific tests — out of scope for this CLI-focused issue.

## Test plan

- [x] New regression tests in `crates/beamtalk-cli/tests/cli_build.rs`: cross-file internal-class leak and internal-type-alias leak both fail `beamtalk build` with E0402
- [x] New unit tests for `ClassHierarchy::stamp_package_on_infos` in `crates/beamtalk-core/src/semantic_analysis/class_hierarchy/tests.rs`
- [x] `just test` — full suite (Rust, stdlib, BUnit, Erlang runtime) passes
- [x] `just ci-changed` — build, clippy, fmt, corpus check, surface-drift check all pass
- [x] Manually verified E0401 (cross-package), E0402 (cross-file class leak, alias leak), and E0403 (cross-package internal method call) all produce the exact message/hint text now documented

---
_Generated by [Claude Code](https://claude.ai/code/session_0173SypWtHsgx1prqQJgG9jk)_

### DIFF
--- a/crates/beamtalk-cli/src/commands/build.rs
+++ b/crates/beamtalk-cli/src/commands/build.rs
@@ -236,6 +236,12 @@ pub fn build(path: &str, options: &beamtalk_core::CompilerOptions, force: bool) 
         .full_manifest
         .as_ref()
         .is_some_and(|m| !m.dependencies.is_empty());
+    // BT-2920: Set the current package so E0401/E0402 visibility checks
+    // (`check_class_visibility`/`check_alias_leaked_visibility`) actually run
+    // — they're gated on `current_package: Some(_)` and silently emit zero
+    // diagnostics otherwise. `None` for single-file/manifest-less builds,
+    // which have no package boundary to enforce.
+    options.current_package = env.pkg_manifest().map(|m| m.name.clone());
     let options = &options;
 
     let dep_ctx = resolve_and_validate_dependencies(&env, options)?;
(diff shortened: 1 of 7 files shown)


jamesc:
Thanks for the review — addressed each:

- **S1** (double manifest read in `run_lint`): fixed in 2b7bc93 — `run_lint` now parses `beamtalk.toml` exactly once, deriving both `current_package` and `has_package_dependencies` from the same read with a single consistent warning path. `resolve_dep_class_infos` (renamed `merge_dependency_class_infos`) no longer re-parses; its only remaining job is merging resolved dependency classes.
- **S2** (`beamtalk test`'s BUnit compilation not setting `current_package`): agreed it's pre-existing and out of scope here — filed [BT-2922](https://linear.app/beamtalk/issue/BT-2922).
- **N1** (missing blank line before `stamp_package_on_infos`): dismissing — checked the surrounding `impl` block and there are no blank lines between *any* adjacent methods (e.g. the pre-existing `stamp_package`/`extract_class_infos` boundary), so the new method is actually consistent with the file's existing style, not a deviation from it.

---

PR #3040, opened 2026-07-16
Wire AliasRegistry into codegen call sites so alias-named annotations emit user_type references in production (BT-2909)

jamesc:
Linear: https://linear.app/beamtalk/issue/BT-2909/wire-aliasregistry-into-codegen-call-sites-so-alias-named-annotations

## Summary

Follow-up to BT-2900, which added `generate_alias_type_attrs` / `format_alias_type_attributes` / `user_type` reference support to `spec_codegen.rs`, fully unit-tested in isolation but never wired into the module-level codegen drivers — they all passed `None` for the `aliases: Option<&AliasRegistry>` parameter, so none of the named-`-type` / `user_type`-reference behavior was live in production (alias names still fell through to `any()`).

This PR does the real pipeline wiring, carefully, because of a correctness trap discovered while implementing BT-2900: a `-spec`/`-type` referencing an undeclared local type is a hard `erlc` compile error, not just a Dialyzer warning. Wiring `Some(registry)` into the spec-generating calls *without also* emitting the matching named `-type` declarations into the same generated module would make real user code with both a `type X = ...` alias and an alias-typed annotation fail to compile.

## Key changes

- Added `AliasRegistry::from_module_declarations(module: &Module) -> Self`, a non-diagnostic constructor that builds the name → annotation map straight from a module's parsed `type_aliases`. Unlike `register_module`, it does no namespace-collision/duplicate/cycle validation — by the time a module reaches codegen it has already passed semantic analysis, so codegen only needs resolution.
- Wired `Some(&registry)` into the four call sites that previously passed `None`: `actor_codegen.rs`, `value_type_codegen.rs`, `supervisor_codegen.rs` (×2), and `gen_server/native_facade.rs`.
- At each of those same call sites, also appended `generate_alias_type_attrs(&registry)` / `format_alias_type_attributes(...)` into the module's `attributes [...]` list (mirroring how `spec_suffix` / `type_alias_suffix` are already appended), so every class module that could contain a `user_type` reference also declares the corresponding named `-type`.
- Removed the now-stale `#[allow(dead_code)]` on `generate_alias_type_attrs` / `format_alias_type_attributes` in `spec_codegen.rs` now that they have real callers.
- New integration tests in `tests/gen_server.rs` and `tests/supervisor.rs` using the full `generate_module`/`CodegenOptions` harness, covering: actor class-method params, Value-subclass state fields, native-facade class methods, a no-type-aliases no-op case (byte-identical output confirmed for modules with an empty alias registry), and an end-to-end `erlc` compile check (`test_alias_annotated_actor_module_compiles_through_erlc`) that would fail if the `user_type` reference were ever wired without the matching named `-type` declaration.

## Test plan

- [x] `cargo test -p beamtalk-core --lib` — 5168 passed, 0 failed, 1 ignored
- [x] `just build`
- [x] `just clippy`
- [x] `just fmt-check`
- [x] Pre-push hook (full `just ci`, including Dialyzer/spec validation and erlc compile checks) — passed

---
_Generated by [Claude Code](https://claude.ai/code/session_01WJNsES5SasW8xTxA1z5yzw)_


jamesc:
Addressed the review suggestions/nits in 25a3292:

1. **Cross-module aliases fall through to `any()`** — added a short comment at each of the four `AliasRegistry::from_module_declarations` call sites (`actor_codegen.rs`, `value_type_codegen.rs`, `supervisor_codegen.rs` ×2, `gen_server/native_facade.rs`) noting that cross-module aliases are not resolved and still fall through to `any()`, per the constructor's documented contract.
2. **`extract_attr_term/2` regex-injection nit** — added a doc note in `validate_specs.escript` that `AttrKey` is spliced directly into the regex source and must be a compile-time literal (`"'spec'"`/`"'type'"`), never a dynamic/user-supplied string.
3. **`type T` / BT-1156 `t()` collision risk** — documented the potential duplicate-`'type'` collision in `alias_erlang_type_name`'s doc comment, noting that `validate_specs.escript`'s Dialyzer/erlc checks would surface it as a compile error rather than silently resolving it.

All three are documentation-only changes (no behavior change), so no new tests were needed per the "skip tests for documentation updates" guidance. `just test`, `just clippy`, and the pre-push `just ci`-equivalent lint/Dialyzer gate all pass on the pushed commit.

---

PR #3395, opened 2026-08-15
feat(cli): warm/persist shared OTP type-spec cache for fresh remote sessions (BT-2471)

jamesc:
Linear: https://linear.app/beamtalk/issue/BT-2471

## Summary

BT-2470 added a shared, OTP-version-keyed FFI type-spec cache (`<cache>/beamtalk/otp-specs/<otp>-<erts>/`, ADR 0075) that eliminates the cold cost of re-extracting hundreds of OTP `.beam` files — but only once it's warm. Claude-Code-on-the-web containers are cloned fresh and reclaimed after inactivity, so that cache never survives between sessions on its own: every fresh remote session paid the cold extraction cost on the first interactive `beamtalk build`.

This adds a background warming step so the shared cache is warm by the time a session's first build runs, without blocking session startup.

- **`beamtalk warm-otp-cache`** (`crates/beamtalk-cli/src/commands/warm_otp_cache.rs`) — a new hidden, project-agnostic CLI command. It's a thin wrapper around the existing `beamtalk_core::ffi_type_specs::extract_type_specs` (the same function `beamtalk build`/`beamtalk lint` already use), so there's no duplicated cache logic — just a call site that needs no `beamtalk.toml`/project root. Its local (non-shared) cache tier is a fresh `tempfile::TempDir`, so only the shared tier it exists to warm persists.
- **`.claude/hooks/worktree-init.sh`** (the repo's `SessionStart` hook) — backgrounds this command (`nohup ... & disown`) behind `CLAUDE_CODE_REMOTE=true`, right after the existing hex-bridge proxy setup so it inherits that PATH/env. It builds `beamtalk-cli` and the Erlang runtime itself rather than waiting for the session's own `just build` — Cargo/rebar3 key their build caches by content, not by which process triggered the build, so this only reorders work onto idle time, never duplicates it.
- **Docs**: new `docs/development/remote-sessions.md` (linked from `docs/development/README.md`) explaining the `SessionStart` hook and the warming step; ADR 0075's "Cold build cost" section gets a new `(e) BT-2471` bullet.

Note: `.claude/hooks/session-start.sh`, mentioned in the original issue as a possible hook location, turned out to be a generated file copied in from a separate skills repo at session start (it's gitignored — see the comment above it in `.gitignore`), not something this repository's git history tracks. `worktree-init.sh` is the actual repo-tracked `SessionStart` hook (registered in `.claude/settings.json`), so that's where the warming step lives.

## Acceptance criteria

- [x] A fresh session's first `beamtalk build` reads OTP specs from a warm shared cache (zero `.beam` reads) — verified: after running `beamtalk warm-otp-cache` against an empty `BEAMTALK_CACHE_DIR`, building a fresh scaffolded project logged `Type specs loaded from cache (zero .beam files read) modules=403 functions=4240`, with `cache_hits=403 cache_misses=0`.
- [x] Keyed by OTP/ERTS version, re-runs only on a version change — inherited for free from `extract_type_specs`'s existing shared-tier cache-hit path (BT-2470); `beamtalk warm-otp-cache` itself adds no new keying logic.
- [x] No effect on local dev — the warming step in `worktree-init.sh` is gated on `CLAUDE_CODE_REMOTE=true`; local dev's platform cache dir already persists across sessions, and the underlying warm-cache path is a fast no-op regardless (see below).
- [x] Documented in `docs/development/remote-sessions.md` + ADR 0075.

## Validation

Manual cache-cold vs cache-warm timing, with `BEAMTALK_CACHE_DIR` pointed at an empty scratch directory:

```
$ time beamtalk warm-otp-cache   # cold: empty shared cache
OTP type-spec cache warm: 403 modules, 4240 functions
real  0m2.739s

$ time beamtalk warm-otp-cache   # warm: same shared cache, second run
OTP type-spec cache warm: 403 modules, 4240 functions
real  0m0.745s   # no .beam files read, no BEAM node spawned
```

Also ran, all green:
- `just build` (Rust + Erlang runtime + stdlib)
- `just test` — full suite: 1108/147/42/22 Rust unit tests, 233 stdlib `.btscript` tests, 3293 BUnit tests, 5663+1813 Erlang EUnit tests, 520 metamorphic assertions — 0 failures
- `cargo clippy -p beamtalk-cli --all-targets -- -D warnings`, `cargo fmt --all -- --check` — clean
- `shellcheck` on the modified hook script — clean
- Pre-push hook (clippy, fmt, Dialyzer, Elixir checks) — passed on push

---
_Generated by [Claude Code](https://claude.ai/code/session_013U6P81vA2ds5mHzVTg6sPn)_


jamesc:
Addressed the review feedback in aa1fe091:

- **Suggestion (rebar3 race):** confirmed real — the background warmer's `rebar3 compile` and `just build`'s (`build-erlang`) both target `runtime/_build/` with no lock between them, and rebar3 has no Cargo-style built-in guard against concurrent writers. Fixed with `flock` against a shared `runtime/.rebar3-compile.lock`: non-blocking in the background warmer (skips gracefully — `warm-otp-cache` already treats "runtime not built" as non-fatal), blocking in `build-erlang` (must always complete). Split `build-erlang` into `[unix]`/`[windows]` variants since PowerShell has no `flock` and Windows has no concurrent background warmer to guard against.
- **Suggestion (surface-parity.md gap):** added the missing `warm-otp-cache` row, mirroring the existing `doctor` row.
- **Nit (doc comment overstatement):** softened "always exits 0" to point at the one real error path.

---

jamesc:
`check (macos-latest)` failed in 9f3d6d30 — my own regression from the previous fix: `flock` isn't preinstalled on macOS (it's Linux-only, util-linux), and the `[unix]` attribute lumped macOS in with Linux. The background warmer this guards against only ever runs on `CLAUDE_CODE_REMOTE=true` Linux sandboxes, so macOS has no concurrent writer to protect against either — split `build-erlang` into `[linux]` (flock-guarded) and `[macos]` (unguarded, matching the existing `[windows]` variant) in 9f3d6d30. Watching CI on the new push.

---

jamesc:
Both findings addressed in 4721ea65:

- **Suggestion (unenforced Linux assumption):** confirmed real — `worktree-init.sh` only checked `CLAUDE_CODE_REMOTE=true`, no `uname` check, so the Justfile's `[linux]`-only flock guard rested on an assumption asserted in a comment rather than enforced. Added `[[ "$(uname -s)" == "Linux" ]] || exit 0` as the first check in the backgrounded warmer.
- **Nit (overstated PATH/HEX_CDN comment):** confirmed — `PATH` is only exported in the auth-proxy branch, not the cloud-sandbox-only path this warmer actually runs in. Corrected.

---

PR #3414, opened 2026-08-18
Fix class-side extension methods never found by class-side dispatch (BT-3192)

jamesc:
## Summary

Linear: https://linear.app/beamtalk/issue/BT-3192

A class-side extension method (`Target class >> sel => body`, ADR 0066 open classes) registered correctly in the `beamtalk_extensions` ETS table, but no class-side dispatch path ever read it back — an ordinary `Target sel` send or explicit `Target class sel` send raised `does_not_understand` instead of reaching the extension.

Root cause: `beamtalk_class_dispatch:handle_class_method_call/6` checked only `LocalClassMethods` and `find_class_method_in_chain/2` (both local/inherited class-method tables) — it never consulted `beamtalk_extensions`. Both `class_method_call` (`Target sel`) and `metaclass_method_call` (`Target class sel`) route through this exact same handler (`beamtalk_object_class:dispatch_class_method/5`), so a single fix covers both message forms.

## Changes

- `beamtalk_class_dispatch:handle_class_method_call/6` now checks the extension registry (keyed under the metaclass tag, `class_object_tag(ClassName)`) **before** the local method table / superclass chain — mirroring `beamtalk_dispatch:lookup/5`'s own "extension checked before local method table" order.
- New `invoke_class_extension/5` helper shares `beamtalk_dispatch:invoke_extension/4`'s calling convention (2-arity value-type funs vs. 3-arity actor funs that thread `ClassVars`), so a class-side extension behaves identically to an instance-side one once dispatch reaches it.
- Exported `beamtalk_dispatch:check_extension/2` so the new code path reuses the existing early-bootstrap ETS guard instead of duplicating it.
- Updated `stdlib/test/extension_sealed_test.bt`'s `testIntegerClassSideExtension` — this test was already wired to lock in the bug (asserting `#not_dispatched` with a comment "When wired up, change to assert: 42") — now asserts the correct dispatched result.
- Added `testClassSideForeignExtensionDispatches` to `stdlib/test/foreign_extension_codegen_test.bt`, exercising the codegen-registered (`String class >> bt2250Banner`) path end-to-end.
- Added 7 new EUnit tests to `beamtalk_class_dispatch_tests.erl`: direct `handle_class_method_call/6` coverage (2-arity fun, 3-arity fun with `ClassVars` threading across calls, extension-priority-over-local-method, `ClassSelf` shape), plus end-to-end coverage via the real `class_send`/`metaclass_method_call` gen_server messages.

## Verification

- Confirmed the fix is load-bearing: reverted the runtime change and re-ran `stdlib/test/extension_sealed_test.bt` — `testIntegerClassSideExtension` failed with `Integer does not understand 'meaningOfLife'`; restoring the fix makes it pass.
- `just test` (Rust, stdlib, BUnit — 3293 tests, runtime EUnit — 5685+1837 tests, metamorphic) all green.
- `just lint` (clippy, rustfmt, Beamtalk/Elixir formatting, Erlang binary-literal lint) and `just dialyzer` both pass.

## Out of scope (filed as follow-up)

- `crates/beamtalk-examples/corpus.json` is stale relative to `docs/beamtalk-language-features.md` on `main` (pre-existing, unrelated to this change — confirmed the regenerated corpus is byte-identical whether or not this PR's diff is applied). Filed as [BT-3197](https://linear.app/beamtalk/issue/BT-3197).
- `respondsTo:`/`includesSelector:` reflection on a class object doesn't reflect *any* class-side method (local or extension) today — a pre-existing, broader gap unrelated to this fix (verified per the issue's own acceptance criteria; not something this change regresses or is required to address).

## Test plan

- [x] `just build`
- [x] `just test`
- [x] `just lint`
- [x] `just dialyzer`
- [x] Manually verified the fix flips `testIntegerClassSideExtension` from failing to passing (and back) by toggling the runtime change

https://claude.ai/code/session_01R75aN2QX3v9RAd2K2XwSss

---
_Generated by [Claude Code](https://claude.ai/code/session_01R75aN2QX3v9RAd2K2XwSss)_


jamesc:
`check-corpus` is red here, but it's failing on `main` too — pre-existing drift unrelated to this PR's diff. Confirmed locally: regenerating `corpus.json` produces a byte-identical result whether or not this PR's changes are applied, and `just check-corpus` already fails on a clean checkout of `main` (`4ade628`). Root cause looks like BT-3186 (#3405) adding a new section to `docs/beamtalk-language-features.md` without regenerating the corpus in the same commit.

Filed [BT-3197](https://linear.app/beamtalk/issue/BT-3197) to fix it on `main`. Will merge `main` into this branch and re-check once that lands, per the base-branch-recovery flow.

---

jamesc:
Good catch — fixed in e640ca9.

`invoke_class_extension/5` now catches and converts a failing extension body instead of letting it propagate, mirroring `apply_class_method_fun/6`'s exact catch structure (script-exit passthrough, NLR relay, `undef` classification, generic catch-all) so a bad extension body becomes a structured `{error, ...}` reply instead of crashing the class's own gen_server:

- Extracted `beamtalk_dispatch:apply_extension_by_arity/4` as the shared "how do I call this fun given its arity" core — `invoke_extension/4` keeps its existing re-raise behavior (correct for instance-side dispatch, which typically runs in the receiver's own process) by wrapping it; `beamtalk_class_dispatch` gets a non-raising building block for class-side crash-safety on top of.
- Added `apply_class_extension_fun/5` + updated `invoke_class_extension/6` accordingly.
- New EUnit regression test (`test_class_extension_crash_does_not_kill_class_process`) proving the class gen_server survives a crashing extension and a follow-up call still succeeds.

Verified with the full `just test` suite (Rust, stdlib, BUnit 3292/3294, runtime EUnit 5686+1837, metamorphic) — all green — plus `just lint` and `just dialyzer`.

On the suggestion about `class_self_dispatch/4`/`class_self_dispatch_local/4` (`self sel` from inside another class method still not checking extensions): agreed it's out of scope here and not a regression. Filed [BT-3198](https://linear.app/beamtalk/issue/BT-3198) as a fast-follow and added a scoping note in `handle_class_method_call/6`'s doc comment.

---

jamesc:
Fixed the stale arity cross-reference in ee90794.

---

jamesc:
Two more fixes pushed:

1. **`check-corpus` fix (actually in-scope this time)**: my earlier claim that the corpus staleness was entirely pre-existing/unrelated was only half right. `main`'s docs-driven drift got fixed independently by #3413 (BT-3190) after I filed BT-3197, and merging `main` in picked that up — but it also surfaced a *second*, genuinely BT-3192-caused drift: `crates/beamtalk-examples/corpus.json` also indexes `stdlib/test/*.bt` files, and I hadn't regenerated it after editing `extension_sealed_test.bt` / `foreign_extension_codegen_test.bt`. Regenerated and committed. `just check-corpus` now passes locally.
2. **Logging-consistency nit** from the last review round: `apply_class_extension_fun/5` now takes `ClassName` as an explicit param (mirroring `apply_class_method_fun/6`'s signature) and logs the plain class name instead of the metaclass tag.

Also reconciled a duplicate "merge main" commit that landed on the remote branch (looked like an automated branch update) with my own local merge — same content, just two different commit objects; merged them together, no conflicts.

Full `just test` suite green (Rust, stdlib, BUnit 3292/3294, runtime EUnit 5686+1837, metamorphic), `just lint` and `just dialyzer` pass, `just check-corpus` now passes too.

---

PR #3517, opened 2026-08-25
Add number-on-the-left coercion benchmark and language docs (BT-3265)

jamesc:
## Summary

Phase 4 of [ADR 0116](https://github.com/jamesc/beamtalk/blob/main/docs/ADR/0116-number-on-the-left-arithmetic-coercion.md) (double-dispatch coercion for number-on-the-left arithmetic), implementing [BT-3265](https://linear.app/beamtalk/issue/BT-3265/benchmark-language-docs-for-number-on-the-left-arithmetic).

BT-3262 added the runtime dispatch helper; BT-3263 wired it into `operators.rs` codegen, including a zero-cost guarantee for statically-typed right operands. This PR turns the ADR's own ad hoc de-risking-spike measurement (a hand-written, standalone `.core` module) into a permanent, re-runnable regression benchmark against the *real* codegen output, and documents the `plusFromNumber:`/`minusFromNumber:`/`timesFromNumber:`/`divFromNumber:` convention for end users.

## Key changes

- **`runtime/perf/bench_collect_selfhost.escript`**: new `bench_number_coercion/0` and `bench_number_coercion_dispatch/0`, mirroring `bench_guard/0`'s own methodology (same tight-loop shape, `N`/`Reps`, `min_us` best-of sampling), but against the actual `try`/`catch`/`is_number`/`send_number_coercion` shape BT-3263 generates (verified against `tests/expressions.rs`), not the spike's hand-written approximation:
  - `total + delta`-shaped case (right operand statically numeric): confirmed **statistically identical** to the pre-ADR bare BIF loop (~0.2% variance across two runs) — the zero-cost structural guarantee, checked empirically.
  - `5 + aVector`-shaped case, happy path (right operand's type genuinely unknown, but numeric at runtime — try succeeds, catch never fires): measures the try/catch/is_number cost, found comparable to the already-accepted `is_number` guard's cost measured in the same run (~7.2× vs ~6.75×) — matching the ADR's own conclusion.
  - Dispatch-triggering reference point: uses the real `beamtalk_message_dispatch:send_number_coercion/4` (not a stub) with the ADR's own REPL example (`5 + "not a vector"`) to characterize the DNU+hint path's cost — ~12 µs/op, paid only on the already-failing path.
- **`docs/development/benchmarks.md`**: new subsection with recorded numbers, following the existing "Arithmetic operator guard vs bare BIF (BT-2709)" section's format and its "(run-dependent)" framing. This sandbox's re-measurement of the *unmodified* `bench_guard/0` landed at ~6.75×, well outside that section's recorded ~2.7–3.0× range — the same "run-dependent" finding the ADR's own spike reported on its own sandbox, called out explicitly rather than silently reconciled.
- **`docs/beamtalk-language-features.md`**: new "Number-on-the-left coercion (`plusFromNumber:` etc.)" subsection under Additive operators (structural precedent: `§ Equality operators cannot be overridden`, the nearby operator-semantics callout). Covers the trigger condition (`5 + aVector`), the reflected-method table, a worked non-commutative example (`minusFromNumber:` computing `n - self`, not `self - n`), the "implement it anyway, to reject with intent" pattern for operators that don't apply to a type, and the DNU-hinting behavior for a missing hook. Cross-references ADR 0116.

## Scope note

Per CLAUDE.md's surface-parity rule: **no `docs/development/surface-parity.md` update needed.** This is a language-semantics change reachable identically from every surface (CLI, REPL, MCP, LSP) via ordinary message sends — not a new per-surface operation.

## Test plan

- [x] `just test-stdlib` — 233/233 passed
- [x] `escript perf/bench_collect_selfhost.escript` — run twice from `runtime/`, numbers agree within ~1%
- [x] `cargo clippy --workspace --all-targets` — clean (no Rust files touched)
- [x] Pre-push hook (full CI: clippy, rustfmt, dialyzer, erlfmt, spec validation, JS/Elixir/Beamtalk lint) — all clean

https://claude.ai/code/session_01VDp1s9rbz3fas625DEV6Y9

---
_Generated by [Claude Code](https://claude.ai/code/session_01VDp1s9rbz3fas625DEV6Y9)_

### DIFF
--- a/crates/beamtalk-examples/corpus.json
+++ b/crates/beamtalk-examples/corpus.json
@@ -51,7 +51,52 @@
       "explanation": "Code example from docs/beamtalk-language-features.md"
     },
     {
-      "id": "docs-beamtalk-language-features-block-103",
+      "id": "docs-beamtalk-language-features-block-100",
+      "title": "Error Messages (beamtalk-language-features)",
+      "category": "language-reference",
+      "tags": [
+        "assign",
+        "assignment",
+        "beamtalk-language-features",
+        "for each",
+        "for-each",
+        "forEach",
+        "iterate",
+        "iteration",
+        "loop",
+        "mutate",
+        "mutation",
+        "set"
+      ],
+      "source": "// This won't compile — stored closure can't mutate fields\nmyBlock := [:item | self.sum := self.sum + item]\nitems do: myBlock",
+      "explanation": "Code example from docs/beamtalk-language-features.md"
+    },
+    {
+      "id": "docs-beamtalk-language-features-block-104",
+      "title": "Custom Timeouts (beamtalk-language-features)",
+      "category": "language-reference",
+      "tags": [
+        "anonymous function",
+        "assign",
+        "assignment",
+        "async",
+        "beamtalk-language-features",
+        "callback",
+        "closure",
+        "concurrent",
+        "gen_server",
+        "genserver",
+        "lambda",
+        "mutate",
+        "mutation",
+        "process",
+        "set"
+      ],
+      "source": "// Wrap an actor with a custom timeout (milliseconds)\nslowDb := db withTimeout: 30000\nslowDb query: sql              // forwarded with 30s timeout\nslowDb stop                    // stop the proxy when done\n\n// Infinite timeout (use with care — blocks indefinitely)\ninfDb := db withTimeout: #infinity\ninfDb query: sql\ninfDb stop",
+      "explanation": "Code example from docs/beamtalk-language-features.md"
+    },
+    {
+      "id": "docs-beamtalk-language-features-block-108",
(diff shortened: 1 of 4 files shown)


jamesc:
`check-corpus` failed because the new language-docs section adds Beamtalk code examples that `build-corpus` picks up from `docs/beamtalk-language-features.md`, growing the corpus from 1131 to 1133 entries — the checked-in `corpus.json` wasn't regenerated for that. Fixed in 1d0e24a (`just build-corpus` + commit); `just check-corpus` now passes locally.

---

jamesc:
Good catch — added the bare/coerced equality assertion in b18bb53, matching `run_size`/`bench_reduce`/`bench_guard_fold`'s convention. Re-ran the escript, output unchanged and no mismatch raised.

---

